PR #122, opened 2023-10-06
add maker side to trigger orders

test9955667:
Add maker option to trigger orders in invoker

### DIFF
--- a/packages/perennial-extensions/contracts/MultiInvoker.sol
+++ b/packages/perennial-extensions/contracts/MultiInvoker.sol
@@ -462,7 +462,7 @@ contract MultiInvoker is IMultiInvoker, Kept {
     function _placeOrder(address account, IMarket market, TriggerOrder memory order) internal isMarketInstance(market) {
         if (order.fee.isZero()) revert MultiInvokerInvalidOrderError();
         if (order.comparison != -1 && order.comparison != 1) revert MultiInvokerInvalidOrderError();
-        if (order.side != 1 && order.side != 2) revert MultiInvokerInvalidOrderError();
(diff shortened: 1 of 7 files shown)


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -462,7 +462,7 @@ contract MultiInvoker is IMultiInvoker, Kept {
     function _placeOrder(address account, IMarket market, TriggerOrder memory order) internal isMarketInstance(market) {
         if (order.fee.isZero()) revert MultiInvokerInvalidOrderError();
         if (order.comparison != -1 && order.comparison != 1) revert MultiInvokerInvalidOrderError();
-        if (order.side != 1 && order.side != 2) revert MultiInvokerInvalidOrderError();
+        if (order.side > 2) revert MultiInvokerInvalidOrderError();
```
arjun-io, line 465: Do we want to avoid using `0` (default value) for these values?

kbrizzle: I think its ok cuz enum-style values (like this) use 0. (it was originally supposed to 0 before we removed it)

packages/perennial-extensions/contracts/types/TriggerOrder.sol
```diff
@@ -36,6 +36,8 @@ library TriggerOrderLib {
     }
 
     function execute(TriggerOrder memory self, Position memory currentPosition) internal pure {
+        if(self.side == 0)
```
kbrizzle, line 39: nit: space after if

arjun-io (approved):

PR #124, opened 2023-10-08
Add closable magic value

kbrizzle:
Adds magic value: `UFixed6Lib.MAX - 1` to fully close positions.

### Notes
- Achieves this by refactoring the pending position context loading to happen prior to update instead of inside the invariant function.
- Results in significant code size improvements due to code deduplication.

### DIFF
--- a/packages/perennial/contracts/Market.sol
+++ b/packages/perennial/contracts/Market.sol
@@ -590,12 +590,12 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         ) revert MarketExceedsPendingIdLimitError();
 
         if (
-            !context.maxPendingPosition.maintained(context.latestVersion, context.riskParameter, context.pendingCollateral)
-        ) revert MarketInsufficientMaintenanceError();
-
-        if (
             !context.currentPosition.local.margined(context.latestVersion, context.riskParameter, context.pendingCollateral)
         ) revert MarketInsufficientMarginError();
+
+        if (
+            !context.maxPendingPosition.maintained(context.latestVersion, context.riskParameter, context.pendingCollateral)
+        ) revert MarketInsufficientMaintenanceError();
 
         if (
             (context.local.protection > context.latestPosition.local.timestamp) &&
--- a/packages/perennial/contracts/Market.sol
+++ b/packages/perennial/contracts/Market.sol
@@ -219,40 +219,41 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     /// @notice Loads the specified global pending position from state and adjusts it
     /// @param context The context to use
     /// @param id The position id to load
-    /// @return position The loaded and global adjusted position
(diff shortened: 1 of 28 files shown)


packages/perennial/contracts/Market.sol
```diff
@@ -10,6 +10,10 @@ import "./interfaces/IMarketFactory.sol";
 /// @notice Manages logic and state for a single market.
 /// @dev Cloned by the Factory contract to launch new markets.
 contract Market is IMarket, Instance, ReentrancyGuard {
+    Fixed6 private constant MAGIC_VALUE_WITHDRAW_ALL_COLLATERAL = Fixed6.wrap(type(int256).min);
```
arjun-io, line 13: Does it hurt to make these public so other clients can read them and pass them in?

packages/perennial/contracts/types/Order.sol
```diff
@@ -143,4 +144,12 @@ library OrderLib {
     function isEmpty(Order memory self) internal pure returns (bool) {
         return self.maker.abs().add(self.long.abs()).add(self.short.abs()).isZero();
     }
+
+    /// @notice Returns the amount of the non-zero side of the order
+    /// @dev Assumes the order must be single-sided
+    /// @param self The Order object to check
+    /// @return The magnitude of the order
+    function magnitude(Order memory self) internal pure returns (Fixed6) {
+        return self.maker.add(self.long).add(self.short);
```
arjun-io, line 153: Should we `abs` here instead of in the caller? 

We can also use this above in `isEmpty`

packages/perennial/contracts/types/Order.sol
```diff
@@ -125,10 +126,10 @@ library OrderLib {
         OracleVersion memory latestVersion,
         RiskParameter memory riskParameter
     ) internal pure returns (UFixed6) {
-        UFixed6 magnitude = self.maker.abs().add(self.long.abs()).add(self.short.abs());
-        if (magnitude.isZero()) return UFixed6Lib.ZERO;
+        if (magnitude(self).isZero()) return UFixed6Lib.ZERO;
```
arjun-io, line 129: `if (isEmpty(self))` ?

packages/perennial/contracts/Market.sol
kbrizzle: Yeah it adds about 120 bytes of code size, so probably not worth it.

packages/perennial/contracts/types/Order.sol
kbrizzle: nice these two save about 40 bytes of code size.

packages/perennial/contracts/types/Order.sol
kbrizzle: This magnitude is signed, unlike the position magnitude.

arjun-io (approved):

PR #125, opened 2023-10-11
Trigger order test cleanup

test9955667:
clean up trigger order tests to make the encoding simple and the args in the tests explicit


packages/perennial-extensions/test/helpers/types.ts
```diff
@@ -46,30 +57,29 @@ export const openTriggerOrder = ({
...
   feePct,
 }: {
   size: BigNumberish
   price: BigNumberish
-  side?: Dir
-  trigger?: TriggerType
+  side: Dir | number
+  comparison: Compare | number
+  orderType: OrderType
   feePct?: BigNumberish
```
arjun-io, line 69: Let's remove the `feePct` as well, this should just be a flat fee amount

arjun-io (approved):

PR #126, opened 2023-10-11
Pyth soft revert burn

test9955667:
problem: soft reverts for pyth don't send back the pyth eth fee in the invoker


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -322,7 +322,10 @@ contract MultiInvoker is IMultiInvoker, Kept {
             IPythOracle(oracleProvider).commit{value: value}(version, data);
         } else {
             try IPythOracle(oracleProvider).commit{value: value}(version, data) { }
-            catch { }
+            catch {
+                // returns the pyth fee sent here on soft-revert
+                payable(msg.sender).transfer(value);
```
arjun-io, line 327: I think we might need to do a `min(msg.value, value)` since theoretically they could "refund" an amount that was never sent in the first place

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -312,18 +312,20 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
         if (revertOnFailure) {
-            IPythOracle(oracleProvider).commit{value: value}(index, version, data);
+            IPythOracle(oracleProvider).commit{value: value}(version, data);
         } else {
-            try IPythOracle(oracleProvider).commit{value: value}(index, version, data) { }
-            catch { }
+            try IPythOracle(oracleProvider).commit{value: value}(version, data) { }
+            catch {
+                // returns the pyth fee sent here on soft-revert
+                payable(msg.sender).transfer(msg.value < value ? msg.value : value);
```
arjun-io, line 327: Can we use `Math.min` here?

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -335,8 +335,12 @@ contract MultiInvoker is IMultiInvoker, Kept {
         if (revertOnFailure) {
             IPythOracle(oracleProvider).commit{value: value}(version, data);
         } else {
-            try IPythOracle(oracleProvider).commit{value: value}(version, data) { }
-            catch { }
+            try IPythOracle(oracleProvider).commit{value: value}(version, data) { } // solhint-disable-line no-empty-blocks
+            catch {
+                // returns the pyth fee sent here and avoids DSU push on soft-revert
+                payable(msg.sender).transfer(Math.min(msg.value, value));
```
kbrizzle, line 341: I'm thinking it might not be worth added in the `Math.min` here.

Even with this you could still more or less arbitrarily exfiltrate ETH by calling multiple `PRICE_COMMIT` actions in a single invocation. It might just be better to take the position that there should be no funds in the MI at rest, and if there are, they can be taken.

arjun-io: That might be a bit too much complexity for this. I agree with Kevin that we should just explicitly state that the multiinvoker cannot custody any ETH safely

test9955667: this achieves the same and justifies the comment, just moves the transfer of msg.value to the end of invoke. will do pyth value transfer in catch tho

test9955667: yeah im down to edit the comments around that. this way, you cant accidentally leave dust from invoke or have  msg.value > sum of pyth commit values

kbrizzle (approved):

PR #128, opened 2023-10-11
Soft revert liq + exec order

test9955667:
Adds `revertOnFailure` flag to liquidate and execute order
Cleans up soft revert try/catch from:
```
if (revertOnFailure) update(
…
) 
else try update(
…
) catch {}
```
to:
```
try update (
…
) catch (bytes memory reason) {if (revertOnFailure) revert (string(reason))} 
```

⚠️ ~~hardhat is able to recognize the revert string as a **custom error from the market** instead of a string error from the invoker... except for [here](https://github.com/equilibria-xyz/perennial-v2/blob/26f5dfd1619557e2b38021beef8a783ce0c7f228/packages/perennial-extensions/test/integration/core/Liquidate.test.ts#L207) where it only works when expecting a revert with panic. ~~IF market is supposed to be throwing a custom error here but it cant be parsed, then we may have to use the old try/catch flow~~ . This revert is a panic, but it may not be clear what the panic is or which contract its coming from~~

Using first try/catch because solidity doesn't sufficiently support try/catch custom errors


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -222,20 +222,31 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
 
-        market.update(
-            account,
-            currentPosition.maker.isZero() ? UFixed6Lib.ZERO : currentPosition.maker.sub(closable),
-            currentPosition.long.isZero() ? UFixed6Lib.ZERO : currentPosition.long.sub(closable),
-            currentPosition.short.isZero() ? UFixed6Lib.ZERO : currentPosition.short.sub(closable),
-            Fixed6Lib.from(-1, liquidationFee),
-            true
-        );
+        if(revertOnFailure) {
```
arjun-io, line 230: nit: Space after `if`

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -222,20 +222,31 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
+            );
+        } else {
+            try market.update(
+                    account,
+                    currentPosition.maker.isZero() ? UFixed6Lib.ZERO : currentPosition.maker.sub(closable),
+                    currentPosition.long.isZero() ? UFixed6Lib.ZERO : currentPosition.long.sub(closable),
+                    currentPosition.short.isZero() ? UFixed6Lib.ZERO : currentPosition.short.sub(closable),
+                    Fixed6Lib.from(-1, liquidationFee),
+                    true
+            // solhint-disable-next-line
```
arjun-io, line 247: Specify the rule you want to disable: `// solhint-disable-next-line <rule name>`

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -322,10 +333,13 @@ contract MultiInvoker is IMultiInvoker, Kept {
         if (revertOnFailure) {
             IPythOracle(oracleProvider).commit{value: value}(index, version, data);
         } else {
-            try IPythOracle(oracleProvider).commit{value: value}(index, version, data) { }
-            catch { }
+            try IPythOracle(oracleProvider).commit{value: value}(index, version, data) { } // solhint-disable-line
```
arjun-io, line 336: Is this the same change as #126 If so we should revert it here.

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -419,22 +434,35 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
+            );
+        } else {
+            try market.update(
+                account,
+                currentPosition.maker,
+                currentPosition.long,
+                currentPosition.short,
+                Fixed6Lib.ZERO,
+                false
+            // solhint-disable-next-line
```
arjun-io, line 460: Specify solhint rule here too

packages/perennial-extensions/test/integration/core/Orders.test.ts
```diff
@@ -375,6 +381,50 @@ describe('Orders', () => {
...
+      multiInvoker
+        .connect(user)
+        .invoke(buildExecOrder({ user: user.address, market: market.address, orderId: 1, revertOnFailure: true })),
+    ).to.be.reverted
+
+    await expect(
+      await multiInvoker
+        .connect(user)
+        .invoke(buildExecOrder({ user: user.address, market: market.address, orderId: 1, revertOnFailure: false })),
+    ).to.not.be.reverted
```
arjun-io, line 425: Can you re-add the collateral and then assert that the same order is correctly executed? This ensures we correctly don't delete the order on soft reverts

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -222,20 +222,31 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
-        market.update(
-            account,
-            currentPosition.maker.isZero() ? UFixed6Lib.ZERO : currentPosition.maker.sub(closable),
-            currentPosition.long.isZero() ? UFixed6Lib.ZERO : currentPosition.long.sub(closable),
-            currentPosition.short.isZero() ? UFixed6Lib.ZERO : currentPosition.short.sub(closable),
-            Fixed6Lib.from(-1, liquidationFee),
-            true
-        );
+        if (revertOnFailure) {
+            market.update(
```
arjun-io, line 231: Pull this update into a separate function and have both paths call it

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -418,22 +432,35 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
-        market.update(
-            account,
-            currentPosition.maker,
-            currentPosition.long,
-            currentPosition.short,
-            Fixed6Lib.ZERO,
-            false
-        );
+        if (revertOnFailure) {
+            market.update(
```
arjun-io, line 442: Move to separate function

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -321,10 +332,12 @@ contract MultiInvoker is IMultiInvoker, Kept {
         if (revertOnFailure) {
             IPythOracle(oracleProvider).commit{value: value}(version, data);
         } else {
-            try IPythOracle(oracleProvider).commit{value: value}(version, data) { }
-            catch { }
+            try IPythOracle(oracleProvider).commit{value: value}(version, data) { } // solhint-disable-line no-empty-blocks
```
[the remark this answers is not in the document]
test9955667: reverted to original of just the pyth call, then merging burn bug pr

packages/perennial-extensions/contracts/MultiInvoker.sol
test9955667: problem: `try` can only be used on external calls, so the separate function would have to contain both paths as it does here to handle `revertOnFailure`

arjun-io: Ah that's annoying...Maybe we can save the position sizes and liq fee to locals and just pass them into the update? I think the update below doesn't need another function

test9955667: im still pro calling update once and bubbling up the error on revert given that hardhat was able to recognize the error and the contract the error came from in every case

packages/perennial-extensions/contracts/interfaces/IMultiInvoker.sol
```diff
@@ -56,6 +56,8 @@ interface IMultiInvoker {
     error MultiInvokerInvalidOrderError();
     // sig: 0x6f462962
     error MultiInvokerCantExecuteError();
+    // sig: 0x1425ea42
+    error FailedInnerCall();
```
kbrizzle, line 60: nit: remove

kbrizzle (approved):

PR #130, opened 2023-10-12
Charge fee on update

test9955667:
consolidates `CHARGE_FEE` action into `UPDATE`

### DIFF
--- a/packages/perennial-extensions/contracts/MultiInvoker.sol
+++ b/packages/perennial-extensions/contracts/MultiInvoker.sol
@@ -250,7 +250,7 @@ contract MultiInvoker is IMultiInvoker, Kept {
         DSU.approve(target);
     }
 
-    /// @notice Chargesa an interface fee to a receiver
+    /// @notice Charges an interface fee to a receiver
     /// @param to Address to receive the fee
     /// @param amount Amount of USDC to transfer
     /// @param wrap Flag to wrap USDC to DSU
--- a/packages/perennial-extensions/contracts/MultiInvoker.sol
+++ b/packages/perennial-extensions/contracts/MultiInvoker.sol
@@ -18,6 +18,11 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @dev Gas buffer estimating remaining execution gas to include in fee to cover further instructions
     uint256 public constant GAS_BUFFER = 100000; // solhint-disable-line var-name-mixedcase
 
+    Fixed6 public constant PERCENT_BASE = Fixed6.wrap(100);
+
+    /// @dev The maximum percentage of collateral an interface can charge to update a position
+    Fixed6 public constant MAX_INTERFACE_FEE_PCT = Fixed6Lib.ONE;
+
     /// @dev USDC stablecoin address
     Token6 public immutable USDC; // solhint-disable-line var-name-mixedcase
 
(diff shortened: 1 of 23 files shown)


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -251,6 +257,32 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
+    /// @param wrap Flag to wrap USDC to DSU
+    function _chargeFee(
+        address to,
+        UFixed6 amount,
+        bool wrap,
+        bool onWithdrawal
+    ) internal {
+        if (amount.isZero()) return;
+
+        if (onWithdrawal) {
```
arjun-io, line 272: It might be simpler to just directly deduct this from the collateral balance via an `update`, similar to how `_raiseKeeperFee` works. Then we wouldn't need the extra logic in the withdraw on line 191

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -18,6 +18,11 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @dev Gas buffer estimating remaining execution gas to include in fee to cover further instructions
     uint256 public constant GAS_BUFFER = 100000; // solhint-disable-line var-name-mixedcase
 
+    Fixed6 public constant PERCENT_BASE = Fixed6.wrap(100);
```
arjun-io, line 21: What are these?

test9955667: missed deleting in cleanup

packages/perennial-extensions/contracts/MultiInvoker.sol
test9955667: our spec is to take it from wallet balance on deposit and collateral on withdrawal. might be a cleaner way to implement than this tho

test9955667: changed to skim fee from collateral

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -174,17 +172,26 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
 
         market.update(msg.sender, newMaker, newLong, newShort, collateral, false);
 
+        // collateral is transferred from the market to this address, an optional interface fee is charged from it,
+        // and the rest is sent to the msg.sender
         Fixed6 withdrawAmount = Fixed6Lib.from(Fixed18Lib.from(DSU.balanceOf()).sub(balanceBefore));
-        // collateral is transferred from the market to this address, transfer to msg.sender from here
-        if (!withdrawAmount.isZero()) _withdraw(msg.sender, withdrawAmount.abs(), wrap);
+        if (!withdrawAmount.isZero()) {
+            withdrawAmount = _chargeFee(market, withdrawAmount, feeInfo);
```
arjun-io, line 192: Charge fee at the very end of this function if interface fee is non-0

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -250,19 +257,28 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
-        emit FeeCharged(msg.sender, to, amount, wrap);
+        // NO-OP (0 fee)
+        if (amount.isZero()) return collateral;
+
+        if (wrap) _unwrap(to, UFixed18Lib.from(amount));
+        else DSU.push(to, UFixed18Lib.from(amount));
+
+        // updates collateral to reflect fee for deposit / withdrawal
+        newCollateral = collateral.sub(Fixed6Lib.from(amount));
+        emit FeeCharged(msg.sender, to, amount, wrap, market);
```
arjun-io, line 281: Move market to the first arg

test9955667: every other event is `Event(account, market...)`, so will change to that instead

arjun-io: 👍  thanks for double checking

packages/perennial-extensions/contracts/MultiInvoker.sol
arjun-io: Either is fine for me, no preference

PR #131, opened 2023-10-13
Enable Pyth batch price commits

kbrizzle:
Moves the price commit flow from the Oracle into the Factory.

External calls and price verification now live in the Factory, which then through a privileged call dispatches the parsed and verified (or invalidated) price to the corresponding oracle. This enables multi-oracle batched updates per oracle integration, and reduces the amount that keeper fees are passed around inside the system.


packages/perennial-oracle/contracts/pyth/PythFactory.sol
```diff
@@ -117,4 +153,33 @@ contract PythFactory is IPythFactory, Factory {
...
+    function _parsePrices(
+        bytes32[] memory ids,
+        uint256 version,
+        bytes calldata data
+    ) private returns (Fixed6[] memory prices) {
+        prices = new Fixed6[](ids.length);
+        bytes[] memory datas = new bytes[](1);
+        datas[0] = data;
+
+        PythStructs.PriceFeed[] memory parsedPrices = pyth.parsePriceFeedUpdates{value: msg.value}(
```
arjun-io, line 171: I think it's safer to do the `singleUpdateFeeInWei() * ids.length` here - there could be a case where the keeper is updating multiple versions in a single tx through multiple `commit` calls.

kbrizzle: Each `commit()` call is external in this contract so it would have its own segregated `msg.value`

packages/perennial-oracle/contracts/pyth/PythFactory.sol
```diff
@@ -78,6 +92,36 @@ contract PythFactory is IPythFactory, Factory {
...
+    function commit(bytes32[] memory ids, uint256 version, bytes calldata data) external payable {
+        bool valid = data.length != 0;
+        Fixed6[] memory prices = valid ? _parsePrices(ids, version, data) : new Fixed6[](ids.length);
+
+        for (uint256 i; i < ids.length; i++)
+            if (IPythOracle(address(oracles[ids[i]])).commit(OracleVersion(version, prices[i], valid)))
+                handleKeep(ids[i], version, prices[i]);
+    }
+
+    function handleKeep(bytes32 id, uint256 version, Fixed6 price)
```
arjun-io, line 113: nit: Add natspec and prefix with `_` since this is a private function

packages/perennial-oracle/contracts/pyth/PythFactory.sol
```diff
@@ -78,6 +92,36 @@ contract PythFactory is IPythFactory, Factory {
...
+    /// @param ids The list of price feed ids to commit
+    /// @param version The oracle version to commit
+    /// @param data The update data to commit
+    function commit(bytes32[] memory ids, uint256 version, bytes calldata data) external payable {
+        bool valid = data.length != 0;
+        Fixed6[] memory prices = valid ? _parsePrices(ids, version, data) : new Fixed6[](ids.length);
+
+        for (uint256 i; i < ids.length; i++)
+            if (IPythOracle(address(oracles[ids[i]])).commit(OracleVersion(version, prices[i], valid)))
+                handleKeep(ids[i], version, prices[i]);
```
arjun-io, line 110: Rename this one too

packages/perennial-oracle/contracts/pyth/PythFactory.sol
```diff
@@ -78,6 +92,36 @@ contract PythFactory is IPythFactory, Factory {
...
+    }
+
+    function _handleKeep(bytes32 id, uint256 version, Fixed6 price)
```
arjun-io, line 113: Do we want to natspec this?

arjun-io (approved):

packages/perennial-oracle/contracts/pyth/PythOracle.sol
```diff
@@ -110,88 +83,40 @@ contract PythOracle is IPythOracle, Instance, Kept {
...
-    /// @notice Commits the price to am invalid requested version
-    /// @dev The keeper reward will not take into account the data cost
-    /// @param version The oracle version to commit
-    /// @param data The update data to commit
-    function _commitInvalidRequested(uint256 version, bytes calldata data)
-        private
-        keep(KEEPER_REWARD_PREMIUM, KEEPER_BUFFER, "", "")
-    {
-        _prices[version] = Fixed6Lib.ZERO;
+        return true;
```
arjun-io, line 108: I just realized, we're missing the `OracleProviderVersionFulfilled` event here

PR #133, opened 2023-10-17
Collateral withdrawal trigger order

test9955667:
Adds a 4th side to trigger orders to support collateral withdrawals


test9955667:
@arjun-io @kbrizzle should this support collateral deposits as well?

PR #135, opened 2023-10-18
[v2.1] Add collateral withdrawal orders

arjun-io:
Adds support for Collateral withdrawal trigger orders

Note: Errors if the order is attempting to increase collateral. This isn't strictly required, we could theoretically have increase collateral orders which would pull from the user's balance, but this is somewhat unintuitive for the user


packages/perennial-extensions/contracts/types/TriggerOrder.sol
```diff
@@ -42,6 +42,10 @@ library TriggerOrderLib {
             currentPosition.long = UFixed6Lib.from(Fixed6Lib.from(currentPosition.long).add(self.delta));
         if (self.side == 2)
             currentPosition.short = UFixed6Lib.from(Fixed6Lib.from(currentPosition.short).add(self.delta));
+        if (self.side == 3)
+            currentPosition.collateral = self.delta; // Collateral updates are already a delta
+        else
+            currentPosition.collateral = Fixed6Lib.ZERO;
```
test9955667, line 48: can this no-op trigger if side reverts over 3?

test9955667 (approved):

kbrizzle (approved):

test9955667 (approved):

PR #141, opened 2023-10-20
[v2.1] Trigger Order interface fee

arjun-io:
~Creates a new mapping within the MultiInvoke to hold TriggerOrderMeta. Currently, this meta struct only holds the interface fee but in the future it can hold more values depending on how complex the order types might get. I've allocated 2 storage slots for this, but we can scale that up or down as needed~

Now just extends the existing struct mapping


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -460,7 +473,8 @@ contract MultiInvoker is IMultiInvoker, Kept {
             currentPosition.collateral,
             false
         ) {
-            delete _orders[account][market][nonce];
+            _chargeFee(market, orderMetas(account, market, nonce).interfaceFee);
```
kbrizzle, line 476: are there any weird edge cases where withdrawing all / too much will cause the interface fee to revert? I suppose you should just not do that, but wondering if there's any time-delay implications there

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -489,7 +503,9 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @param account Account to place order for
     /// @param market Market to place order in
     /// @param order Order state to place
-    function _placeOrder(address account, IMarket market, TriggerOrder memory order) internal isMarketInstance(market) {
+    function _placeOrder(
+        address account, IMarket market, TriggerOrder memory order, InterfaceFee memory interfaceFee
```
kbrizzle, line 507: nit: newline these

packages/perennial-extensions/contracts/types/TriggerOrderMeta.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+
+import "../interfaces/IMultiInvoker.sol";
+import "@equilibria/root/number/types/UFixed6.sol";
+import "@equilibria/perennial-v2/contracts/types/Position.sol";
+
+struct TriggerOrderMeta {
+    IMultiInvoker.InterfaceFee interfaceFee;
```
kbrizzle, line 9: It might be a little clean to pull `IMultiInvoker.InterfaceFee` into its own memory-only type like `Order` since we're using it now in a few places

packages/perennial-extensions/contracts/types/TriggerOrderMeta.sol
```diff
@@ -0,0 +1,55 @@
...
+}
+using TriggerOrderMetaLib for TriggerOrderMeta global;
+struct StoredTriggerOrderMeta {
+    /* slot 0 */
+    uint40 interfaceFee;// <= 1M USDC
```
kbrizzle, line 14: I think fees in Perennial are `uint48` so might just wanna match that.

packages/perennial-extensions/contracts/types/TriggerOrderMeta.sol
```diff
@@ -0,0 +1,55 @@
...
+    address interfaceFeeReceiver;
+    bool interfaceFeeUnwrap;
+    bytes6 __unallocated0__;
+
+    /* slot 1 */
+    bytes32 __unallocated1__;
```
kbrizzle, line 20: since the original `TriggerOrder` is the value in a mapping, I think it's actually safe to extend that instead of making a new type here -- just thinking through that since this extra slot for future use you've added here I don't think is necessary. Worth looking back over how storage works with structs and mappings.

packages/perennial-extensions/contracts/MultiInvoker.sol
arjun-io: Yeah I just thought of this as well - I don't think it affects the interface fee since those should generally only be charged on position changes where collateral is available in the account, but we may need a separate special case to handle full collateral withdrawals with order keeper fees.

packages/perennial-extensions/contracts/types/TriggerOrderMeta.sol
arjun-io: Yeah I believe you're correct, we can just extend the existing type since the memory location of the struct is not up against other members in the mapping

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -510,6 +516,10 @@ contract MultiInvoker is IMultiInvoker, Kept {
         emit OrderCancelled(account, market, nonce);
     }
 
+    function _deleteOrder(address account, IMarket market, uint256 nonce) internal {
```
test9955667, line 519: this is unused

packages/perennial-extensions/contracts/types/TriggerOrder.sol
```diff
@@ -3,23 +3,30 @@ pragma solidity ^0.8.13;
 
 import "@equilibria/root/number/types/UFixed6.sol";
 import "@equilibria/perennial-v2/contracts/types/Position.sol";
+import "./InterfaceFee.sol";
```
test9955667, line 6: might not make sense to have `InterfaceFee` definition in a diff types file if its fields are being stored /packed uniquely in `StoredTriggerOrder`

test9955667: eh also breaks convention to have 2 types declared in the same file

packages/perennial-extensions/contracts/types/InterfaceFee.sol
```diff
@@ -0,0 +1,15 @@
...
+struct InterfaceFee {
+    /// @dev The amount of the fee
+    UFixed6 amount;
+
+    /// @dev The address to send the fee to
+    address receiver;
+
+    /// @dev Whether or not to unwrap the fee
+    bool unwrap;
+}
```
kbrizzle, line 15: ni: newline

packages/perennial-extensions/contracts/types/TriggerOrder.sol
kbrizzle: We do this pattern elsewhere like for `JumpRateUtilizationCurve` / `Order` where we have these memory-only types that may be used in multiple places. Then when it comes time to store them, each storage type implements custom logic to read / store it.

kbrizzle (approved):

packages/perennial-extensions/contracts/types/TriggerOrder.sol
kbrizzle: We do this pattern elsewhere like for JumpRateUtilizationCurve / Order where we have these memory-only types that may be used in multiple places. Then when it comes time to store them, each storage type implements custom logic to read / store it.

PR #142, opened 2023-10-20
[v2.1][Fix] Fix Collateral withdraw orders

arjun-io:
Fixes 3 issues with Collateral Withdrawal orders:
1. Collateral was not pushed to the account
1. Max Collateral withdrawal orders were not possible due to the magic value being larger in width than the delta storage
1. Max Collateral withdrawal orders were not possible due to the order fee being debited from collateral balance after full withdrawal

Fixes 1 by reusing the `_update` logic for both update and execute actions
Fixes 2 by mapping min int64 to max withdraw for orders
Fixes 3 by executing the max withdraw after the fee debit in `_raiseKeeperFee` - this is somewhat ugly but I'm not sure if there's a cleaner workaround here.


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -485,6 +493,22 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
+                market,
+                UFixed6Lib.MAX,
+                UFixed6Lib.MAX,
+                UFixed6Lib.MAX,
+                Fixed6.wrap(type(int256).min),
+                true,
+                order.interfaceFee,
+                revertOnFailure
+            );
+            delete _orders[account][market][nonce];
```
test9955667, line 509: unused `_deleteOrder` [here](https://github.com/equilibria-xyz/perennial-v2/pull/141/commits/82c1128c6b01ca7c6a316254d7230027f50d0cca) could be used. code only repeated twice tho

packages/perennial-extensions/contracts/types/TriggerOrder.sol
```diff
@@ -52,7 +53,10 @@ library TriggerOrderLib {
             currentPosition.short = UFixed6Lib.from(Fixed6Lib.from(currentPosition.short).add(self.delta));
 
         // update collateral (override collateral field in position since it is not used in this context)
-        currentPosition.collateral = (self.side == 3) ? self.delta : Fixed6Lib.ZERO;
+        // Handles collateral withdrawal magic value
+        currentPosition.collateral = (self.side == 3) ?
+            self.delta.eq(Fixed6.wrap(type(int64).min)) ? Fixed6Lib.MIN : self.delta
```
kbrizzle, line 58: nit: can you put a `()` around the line and include the `:` at the end for consistency and readability here

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -434,57 +435,55 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
             currentPosition.collateral,
-            false
-        ) {
-            _chargeFee(market, orders(account, market, nonce).interfaceFee);
-            delete _orders[account][market][nonce];
-            emit OrderExecuted(account, market, nonce);
-        } catch (bytes memory reason) {
-            if (revertOnFailure) Address.verifyCallResult(false, reason, "");
-        }
+            true,
```
kbrizzle, line 458: why did this go from false to true?

arjun-io: The previous `false` was the protect flag, the new `true` is the `wrap` flag

kbrizzle (approved):

packages/perennial-extensions/contracts/MultiInvoker.sol
kbrizzle: just to confirm my understanding here, we're passing in `true` to have is always send `USDC` back to the user during a `side == 3` withdrawal? and a (wrapped) deposit will never get called here?

PR #158, opened 2023-11-17
Add OrderCreated event

arjun-io:


packages/perennial/contracts/Market.sol
```diff
@@ -301,6 +301,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
 
         // events
         emit Updated(msg.sender, account, context.currentTimestamp, newMaker, newLong, newShort, collateral, protect);
+        emit OrderCreated(account, context.currentTimestamp, collateral, newOrder);
```
kbrizzle, line 304: nit: I'd put `collateral` after `newOrder`, just to match the order that we have those in in `Updated`

kbrizzle (approved):

PR #163, opened 2023-11-25
[Fix] Make fees nonnegative

kbrizzle:
Temporarily resolves: https://github.com/sherlock-audit/2023-10-perennial-judging/issues/58, while we design a more holistic algorithm.


packages/perennial/contracts/types/Order.sol
arjun-io, line 1: Do we also want to revert the `UFixed6 -> Fixed6` change for the `fee` value or keep that in?

kbrizzle: I think we should keep that in since we'll need it in the future and we want to minimize change size in the audit fixes.

arjun-io (approved):

PR #164, opened 2023-11-25
[Fix] Multiple callback registrations during the same version

kbrizzle:
Resolves https://github.com/sherlock-audit/2023-10-perennial-judging/issues/25.


arjun-io (approved):
Changes LGTM but looks like there's some failing tests in Extensions, might be unrelated?

PR #166, opened 2023-11-27
[Fix] Order eligibility w/ invalid versions

kbrizzle:
Resolves: https://github.com/sherlock-audit/2023-10-perennial-judging/issues/31.

### DIFF
--- a/packages/perennial-extensions/contracts/types/TriggerOrder.sol
+++ b/packages/perennial-extensions/contracts/types/TriggerOrder.sol
@@ -37,9 +37,10 @@ using TriggerOrderStorageLib for TriggerOrderStorage global;
  * @notice
  */
 library TriggerOrderLib {
-    function fillable(TriggerOrder memory self, Fixed6 latestPrice) internal pure returns (bool) {
-        if (self.comparison == 1) return latestPrice.gte(self.price);
-        if (self.comparison == -1) return latestPrice.lte(self.price);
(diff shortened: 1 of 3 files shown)


packages/perennial-extensions/contracts/types/TriggerOrder.sol
```diff
@@ -37,9 +37,10 @@ using TriggerOrderStorageLib for TriggerOrderStorage global;
  * @notice
  */
 library TriggerOrderLib {
-    function fillable(TriggerOrder memory self, Fixed6 latestPrice) internal pure returns (bool) {
-        if (self.comparison == 1) return latestPrice.gte(self.price);
-        if (self.comparison == -1) return latestPrice.lte(self.price);
+    function fillable(TriggerOrder memory self, OracleVersion memory latestVersion) internal pure returns (bool) {
+        if (!latestVersion.valid) return false;
```
arjun-io, line 41: nit: is it cheaper to do `return latestVersion.valid` at the end instead of the extra branch?

kbrizzle: I tried this and it actually causes a bug because you need to short-circuit first thing if it's invalid

arjun-io (approved):

packages/perennial-extensions/contracts/types/TriggerOrder.sol
arjun-io: ah right that makes sense since the comparison branches can return true early

PR #177, opened 2023-12-02
Chainlink oracle

0xfennel:


packages/perennial-oracle/contracts/chainlink/ChainlinkFactory.sol
```diff
@@ -0,0 +1,89 @@
...
+    /// @param version The oracle version to validate against
+    /// @param data The update data to validate
+    /// @return prices The parsed price list if valid
+    function _parsePrices(
+        bytes32[] memory ids,
+        uint256 version,
+        bytes calldata data
+    ) internal override returns (Fixed6[] memory prices) {
+        if (ids.length != 1) revert ChainlinkFactoryMultipleIdsError();
+        prices = new Fixed6[](ids.length);
```
kbrizzle, line 55: fine to hardcode this to 1 if we're explicitly not enabling batching for now.

packages/perennial-oracle/contracts/chainlink/ChainlinkFactory.sol
```diff
@@ -0,0 +1,89 @@
...
+        (bytes32 feedId, , uint32 observationsTimestamp, , , , uint192 price) =
+            abi.decode(verifiedReport, (bytes32, uint32, uint32, uint192, uint192, uint32, uint192));
+
+        if (
+            version < observationsTimestamp + validFrom ||
+            version > observationsTimestamp + validTo
+        ) revert ChainlinkFactoryVersionOutsideRangeError();
+        if (feedId != toUnderlyingId[ids[0]]) revert ChainlinkFactoryInvalidFeedIdError(feedId);
+
+        prices[0] = Fixed6Lib.from(Fixed18.wrap(SafeCast.toInt256(price)));
```
kbrizzle, line 67: the safecast step isn't necessary in this case since you're going from smaller to larger bits

kbrizzle: not really less steps, but I believe you can do it with just root

packages/perennial-oracle/contracts/chainlink/ChainlinkFactory.sol
```diff
@@ -0,0 +1,89 @@
...
+        internal override
+        keep(
+            commitKeepConfig(numRequested),
+            msg.data[0:0],
+            applicableValue(data),
+            ""
+        )
+    { }
+
+    function applicableValue(bytes calldata data) private returns (uint256) {
```
kbrizzle, line 84: natspec

packages/perennial-oracle/contracts/chainlink/ChainlinkFactory.sol
```diff
@@ -0,0 +1,89 @@
...
+/// @notice Factory contract for creating and managing Chainlink oracles
+contract ChainlinkFactory is IChainlinkFactory, KeeperFactory {
+    /// @dev Chainlink verifier contract
+    IVerifierProxy public immutable chainlink;
+
+    /// @dev Chainlink fee manager contract
+    IFeeManager public immutable feeManager;
+
+    /// @dev Fee token address
+    address feeTokenAddress;
```
kbrizzle, line 18: make `public` and `immutable`

packages/perennial-oracle/contracts/interfaces/IChainlinkFactory.sol
```diff
@@ -0,0 +1,59 @@
...
+
+import "./IKeeperFactory.sol";
+
+interface IChainlinkFactory is IKeeperFactory {
+    error ChainlinkFactoryMultipleIdsError();
+    error ChainlinkFactoryVersionOutsideRangeError();
+    error ChainlinkFactoryInvalidFeedIdError(bytes32 feedId);
+}
+
+interface IVerifierProxy {
```
kbrizzle, line 12: do they not have any sort of lib we can import this from?

kbrizzle: We should give them a heads up on that, it's probably not intended to be pinned since you'd want to import this.

But for now makes sense to keep this in here.

packages/perennial-oracle/contracts/chainlink/ChainlinkFactory.sol
```diff
@@ -0,0 +1,78 @@
...
+            abi.decode(verifiedReport, (bytes32, uint32, uint32, uint192, uint192, uint32, uint192));
+
+        if (
+            observationsTimestamp < version + validFrom ||
+            observationsTimestamp > version + validTo
+        ) revert ChainlinkFactoryVersionOutsideRangeError();
+        if (feedId != toUnderlyingId[ids[0]]) revert ChainlinkFactoryInvalidFeedIdError(feedId);
+
+        prices[0] = Fixed6Lib.from(Fixed18Lib.from(UFixed18.wrap(price)));
+        return prices;
```
kbrizzle, line 67: nit: you don't need to include this line because it's defined in the return value.

PR #182, opened 2023-12-03
[Cleanup] Allocate all of Vault assets

kbrizzle:
Refactors the Vault allocation code to allocate all assets instead of leaving the rounding error as dust balance in the Vault contract.

This allows the Vault to skip reading it's own balance as part of multiple total collateral calculations.


packages/perennial-vault/contracts/lib/StrategyLib.sol
```diff
@@ -143,16 +151,23 @@ library StrategyLib {
...
+        UFixed6 deposit,
+        UFixed6 withdrawal,
+        UFixed6 ineligable
     ) private pure returns (MarketTarget[] memory targets) {
         _AllocateLocals memory _locals;
-        (_locals.totalWeight, _locals.totalMargin) = _aggregate(registrations, strategy.marketContexts);
+        (_locals.totalWeight, _locals.totalMargin, _locals.totalCollateral) =
+            _aggregate(registrations, strategy.marketContexts);
+
+        // TODO: revert if not enough collateral to redeem / get rid of maxRedeem check
```
arjun-io, line 168: Is this going to be a follow-up PR?

arjun-io (approved):

packages/perennial-vault/contracts/lib/StrategyLib.sol
kbrizzle: Yep it's another task that's up on linear

PR #184, opened 2023-12-05
v2.1 codesize

kbrizzle:
#### Background
The reward system has not yet been turned on since launch, so there are no reward balances within the system that need to be kept track of for backwards compatibility. This system takes up considerably code complexity within the protocol, and is a good candidate for removal, since we've determined more efficient ways to achieve the same incentivization goals.

For this first pass, we will only remove the minimum changeset required to statifies our codesize requirements, optimizing for least criticial codepaths first.

- Removes the `reward()` token getter and setter so that the reward token cannot be set.
- Removes the `claimReward()` function since there are no reward, and can no longer be any rewards, in the system.
- Removes the `_accumulateReward()` function within `Version` since there cannot be a non-zero reward rate without a non-zero reward token.

### DIFF
--- a/packages/perennial/contracts/types/Version.sol
+++ b/packages/perennial/contracts/types/Version.sol
@@ -122,10 +122,6 @@ library VersionLib {
         (values.pnlMaker, values.pnlLong, values.pnlShort) =
             _accumulatePNL(self, fromPosition, fromOracleVersion, toOracleVersion);
 
-        // accumulate reward
-        (values.rewardMaker, values.rewardLong, values.rewardShort) =
-            _accumulateReward(self, fromPosition, fromOracleVersion, toOracleVersion, marketParameter);
-
         return (values, values.positionFeeFee.add(values.fundingFee).add(values.interestFee));
     }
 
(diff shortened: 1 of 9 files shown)


packages/perennial/contracts/types/Version.sol
```diff
@@ -293,38 +289,6 @@ library VersionLib {
...
-    /// @notice Globally accumulates position's reward share since last oracle update
-    /// @param self The Version object to update
-    /// @param position The previous latest position
-    /// @param fromOracleVersion The previous latest oracle version
-    /// @param toOracleVersion The next latest oracle version
-    /// @param marketParameter The market parameter
-    /// @return rewardMaker The total reward accrued by makers
-    /// @return rewardLong The total reward accrued by longs
-    /// @return rewardShort The total reward accrued by shorts
-    function _accumulateReward(
```
arjun-io, line 306: Do we actually need to delete this function or does the optimizer already drop it if we never call it?

arjun-io (approved):

PR #190, opened 2023-12-09
[Cleanup] Use new Kept

kbrizzle:
Incorporate changes from: https://github.com/equilibria-xyz/root/pull/84.

### DIFF
--- a/packages/perennial-extensions/contracts/MultiInvoker.sol
+++ b/packages/perennial-extensions/contracts/MultiInvoker.sol
@@ -417,7 +417,7 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @return Amount of keeper fee raised
     function _raiseKeeperFee(UFixed18 keeperFee, bytes memory data) internal virtual override returns (UFixed18) {
         (address account, IMarket market, UFixed6 fee) = abi.decode(data, (address, IMarket, UFixed6));
-        UFixed6 raisedKeeperFee = UFixed6Lib.from(keeperFee.min(UFixed18Lib.from(fee)), true);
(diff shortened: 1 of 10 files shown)


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -377,7 +377,18 @@ contract MultiInvoker is IMultiInvoker, Kept {
 
         TriggerOrder memory order = orders(account, market, nonce);
         // Pay out keeper fee based on static gas buffer
-        _handleKeep(account, market, order.fee);
+        _handleKeeperFee(
```
arjun-io, line 380: Can we delete the `_handleKeep` function?

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -422,18 +433,23 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @notice Helper function to raise keeper fee
     /// @param keeperFee Keeper fee to raise
     /// @param data Data to raise keeper fee with
-    function _raiseKeeperFee(UFixed18 keeperFee, bytes memory data) internal virtual override {
+    /// @return Amount of keeper fee raised
+    function _raiseKeeperFee(UFixed18 keeperFee, bytes memory data) internal virtual override returns (UFixed18) {
         (address account, IMarket market, UFixed6 fee) = abi.decode(data, (address, IMarket, UFixed6));
         if (keeperFee.gt(UFixed18Lib.from(fee))) revert MultiInvokerMaxFeeExceededError();
```
arjun-io, line 439: Should we do min(keeperFee, fee) here instead? That would allow keepers to submit orders where they might be okay with getting less than the full keeper fee

arjun-io (approved):

PR #194, opened 2023-12-15
[Cleanup] Remove unneeded LocalAccumulationResult struct

kbrizzle:


packages/perennial/contracts/Market.sol
```diff
@@ -520,7 +518,9 @@ contract Market is IMarket, Instance, ReentrancyGuard {
             newPosition.timestamp,
             fromId,
             newPositionId,
-            accumulationResult
+            collateralAmount,
```
arjun-io, line 521: Does this save a bunch of gas? It'll be slightly annoying to re-do the way the graph works to handle both versions of this event so if we can still emit the struct for the event that would be ideal

PR #200, opened 2023-12-22
Move payoff to oracle

kbrizzle:
# Background
Moves the payoff specification from the Market into the sub-oracle.
- Allows for arbitrary-decimal oracle feeds
- Reduces complexity of Market and contracts that interact with Market.
- Makes way for further future Oracle simplification

# Migration Notes
- KeeperFactory is not backwards compatible, will need to update in Oracle instead of upgrading the implementation
- Non-Zero Payoff Markets not backwards compatible.
  - Needs fresh oracle deployed since previously was pointed to the Zero Payoff oracle
  - Needs a Switchover Oracle contract written to add the payoff to the previous oracle, while acting as a passthrough to the new oracle until settlement has fully switched over
  - Needs a way to switch Oracles


packages/perennial/contracts/interfaces/IMarket.sol
```diff
@@ -112,7 +110,6 @@ interface IMarket is IInstance {
     function initialize(MarketDefinition calldata definition_) external;
     function token() external view returns (Token18);
     function oracle() external view returns (IOracleProvider);
-    function payoff() external view returns (IPayoffProvider);
```
arjun-io, line 115: It might be safer (more backwards compatible) to keep this function and have it return `address(0)`? That way any integrators would maintain the correct logic if they still have a `transform` hardcoded

We can then fully remove this in `2.3` to give them some time

arjun-io (approved):

PR #208, opened 2024-01-07
[WIP] v2.2 Fee Update

kbrizzle:
#### `PositionGlobal`
  - Slot 0
    - Deprecates `fee` and `keeper`
    - Adds `orders` over top
  - Slot 2
    - Adds `makerPos`, `makerNeg`, `takerPos`, `takerNeg`

Slot 0 changes are not backwards compatible. Pending `fee` and `keeper` at time of upgrade will be skipped, or re-read as `orders`. Must go into no-order mode and settle all non-empty orders prior to update.

Slot 2 is backwards compatible as it defaults to `ZERO` which will not add additional fees to the order when settled.


packages/perennial/contracts/types/Version.sol
```diff
@@ -45,235 +66,384 @@ struct VersionAccumulationResult {
...
 }
 
 ///@title Version
 /// @notice Library that manages global versioned accumulator state.
 /// @dev Manages the value accumulator which measures the change in position value over time.
 library VersionLib {
+    /// @notice Resets the per-version accumulators to prepare for the next version
+    /// @param self The Version object to update
+    function next(Version memory self) internal pure {
+        self.makerPosFee._value = Fixed6Lib.ZERO;
```
arjun-io, line 86: Should we add a `reset` function to accumulators?

packages/perennial/contracts/types/Version.sol
```diff
@@ -45,235 +66,384 @@ struct VersionAccumulationResult {
...
+        UFixed6 orderMagnitude,
+        UFixed6 linearFee,
+        UFixed6 proportionalFee
+    ) private pure {
+        UFixed6 positionFee = linearFee.add(proportionalFee);
+        feeAccumulator.decrement(Fixed6Lib.from(positionFee), orderMagnitude);
+
+        UFixed6 protocolFee = context.fromPosition.maker.isZero() ? positionFee : context.marketParameter.positionFee.mul(positionFee);
+        UFixed6 positionFeeMaker = positionFee.sub(protocolFee);
+        self.makerValue.increment(Fixed6Lib.from(positionFeeMaker), context.fromPosition.maker);
```
arjun-io, line 296: Where are we incrementing the taker value accumulators?

packages/perennial-extensions/contracts/types/TriggerOrder.sol
```diff
@@ -68,9 +72,7 @@ library TriggerOrderLib {
 
         // update collateral (override collateral field in position since it is not used in this context)
```
arjun-io, line 73: We can remove this comment

packages/perennial/contracts/Market.sol
```diff
@@ -112,6 +118,18 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     /// @notice Updates the risk parameter set of the market
     /// @param newRiskParameter The new risk parameter set
     function updateRiskParameter(RiskParameter memory newRiskParameter) external onlyCoordinator {
+
+        // credit impact update fee to the protocol account
+        Global memory latestGlobal = _global.read();
```
arjun-io, line 123: Should we settle before doing this?

packages/perennial/contracts/Market.sol
```diff
@@ -529,33 +595,13 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
-        Position memory newPosition,
-        Version memory version
-    ) private {
-        if (context.local.processProtection(newPosition, version)) {
-            Local memory localInitiator = _locals[context.local.protectionInitiator].read();
-            localInitiator.processLiquidationFee(context.local);
-            _locals[context.local.protectionInitiator].store(localInitiator);
-        }
+    /// @notice Credits an account's collateral that is out-of-context
+    /// @param account The account to credit
```
arjun-io, line 599: Can we add a `dev` note here that this should only be called when the correspond debit is accounted for?

packages/perennial/contracts/Market.sol
kbrizzle: We don't strictly need to as this parameter update will just take place starting at an earlier vs later order id.

packages/perennial/contracts/types/Version.sol
kbrizzle: Added a follow up ticket.

packages/perennial/contracts/types/Version.sol
kbrizzle: This logic is moving the position fee amount from the `feeAccumulator` to the `makerValue` accumulator.

It doesn't touch the taker value accumulator because we're not charging *all taker* the fee, we're just charging the specific fee accumulator that's being charged in this settlement.

arjun-io (approved):
Changes LGTM, we could probably use natspec for the new Version internal functions

PR #218, opened 2024-01-24
Remove Invalidation / Replace pending positions with orders

kbrizzle:


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -351,7 +351,9 @@ contract MultiInvoker is IMultiInvoker, Kept {
 
         _marketSettle(market, account);
 
-        Position memory currentPosition = market.pendingPositions(account, market.locals(account).currentId);
+        Order memory pending = market.pendings(account);
+        Position memory currentPosition = market.positions(account);
+        currentPosition.update(pending, true);
```
arjun-io, line 356: This somewhat reminds me of having to remember to do the adjustment each time which was a little tricky for integrators - maybe we can have a boolean flag in `market.positions` to include pending?

packages/perennial/contracts/Market.sol
```diff
@@ -44,24 +44,33 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
     mapping(uint256 => OrderStorageGlobal) private _pendingOrder;
 
+    /// @dev The local pending order for each id for each account
+    mapping(address => mapping(uint256 => OrderStorageGlobal)) private _pendingOrders;
+
+    /// @dev The global aggregate pending order
+    OrderStorageGlobal private _pending;
+
+    /// @dev The local aggregate pending order for each account
+    mapping(address => OrderStorageGlobal) private _pendings;
```
arjun-io, line 72: Is adding these structs above the checkpoints mapping safe? I guess since this isn't live yet it's okay

packages/perennial/contracts/Market.sol
```diff
@@ -44,24 +44,33 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
     mapping(uint256 => VersionStorage) private _versions;
 
     /// @dev The global pending order for each id
     mapping(uint256 => OrderStorageGlobal) private _pendingOrder;
 
+    /// @dev The local pending order for each id for each account
+    mapping(address => mapping(uint256 => OrderStorageGlobal)) private _pendingOrders;
+
+    /// @dev The global aggregate pending order
+    OrderStorageGlobal private _pending;
```
arjun-io, line 69: We could theoretically save a slot by moving this into the below mapping under `address(this)`

packages/perennial-extensions/contracts/MultiInvoker.sol
kbrizzle: I agree, I made a follow up task for this already.

packages/perennial/contracts/Market.sol
kbrizzle: It is because the Checkpoint storage was added as part of v2.2

packages/perennial/contracts/Market.sol
kbrizzle: I'll add a follow up task to look into this, I'm leaning toward it not being worth it given the issues we had with the vault. But it's worth at least looking into now that we're redoing the pending flow entirely.

arjun-io (approved):

PR #251, opened 2024-02-12
Coordinator Contract

0xfennel:
Tests coming soon, just putting this PR out so that the implementation can be reviewed in the meantime.


packages/perennial-extensions/contracts/Coordinator.sol
```diff
@@ -0,0 +1,58 @@
...
+    /// @dev The address of the fee claimer
+    address public feeClaimer;
+
+    /// @dev The address of the risk parameter updater
+    address public riskParameterUpdater;
+
+    /// @notice Constructs the contract
+    /// @param feeClaimer_ The address of the fee claimer
+    /// @param riskParameterUpdater_ The address of the risk parameter updater
+    constructor(address feeClaimer_, address riskParameterUpdater_) {
```
kbrizzle, line 19: you need to manually initialize the `Ownable` here.

packages/perennial-extensions/contracts/Coordinator.sol
```diff
@@ -0,0 +1,58 @@
...
+    /// @param riskParameter The new risk parameter
+    function updateRiskParameter(IMarket market, RiskParameter calldata riskParameter) external {
+        if (msg.sender != riskParameterUpdater) revert NotRiskParameterUpdater();
+        market.updateRiskParameter(riskParameter);
+    }
+
+    /// @notice Withdraws tokens from the contract
+    /// @param token The token to withdraw the fees for
+    /// @param beneficiary The address to withdraw the fees to
+    function withdraw(Token18 token, address beneficiary) external {
```
kbrizzle, line 54: I think it'd be simpler to just have it withdraw when `claimFee` is called. We always want the claimed funds to go to the `feeClaimer` address.

packages/perennial-extensions/contracts/Coordinator.sol
```diff
@@ -0,0 +1,58 @@
...
+pragma solidity 0.8.19;
+
+import { Ownable } from "@equilibria/root/attribute/Ownable.sol";
+import { ICoordinator, IMarket, RiskParameter, Token18 } from "./interfaces/ICoordinator.sol";
+
+/// @title Coordinator
+/// @notice Manages claiming fees and updating risk parameters for markets
+contract Coordinator is ICoordinator, Ownable {
+    /// @dev The address of the fee claimer
+    address public feeClaimer;
```
kbrizzle, line 11: call these `comptroller` and `coordinator` for the fee claimer and risk param update respectively.

packages/perennial-extensions/contracts/Coordinator.sol
kbrizzle: we can leave these initialized here so the owner has to manually set them, that's a little cleaner.

packages/perennial-extensions/contracts/Coordinator.sol
```diff
@@ -0,0 +1,58 @@
...
+    /// @param riskParameterUpdater_ The address of the risk parameter updater
+    constructor(address feeClaimer_, address riskParameterUpdater_) {
+        feeClaimer = feeClaimer_;
+        riskParameterUpdater = riskParameterUpdater_;
+    }
+
+    /// @notice Updates the fee claimer
+    /// @param feeClaimer_ The address of the new fee claimer
+    function setFeeClaimer(address feeClaimer_) external onlyOwner {
+        feeClaimer = feeClaimer_;
```
kbrizzle, line 27: might be worth sticking an event on these two just for completeness

packages/perennial-extensions/contracts/Coordinator.sol
0xfennel: do you mean the other way around? iirc comptroller is the risk param updater in compound

kbrizzle: comptroller just means the accountant of the system, compound uses this for the name of their central contract that manages all the balances.

kbrizzle (approved):

packages/perennial-extensions/contracts/Coordinator.sol
```diff
@@ -0,0 +1,51 @@
...
+        emit CoordinatorSet(coordinator_);
+    }
+
+    /// @notice Claims the fee for a market
+    /// @param market The market to claim the fee for
+    function claimFee(IMarket market) external {
+        if (msg.sender != comptroller) revert NotComptroller();
+        market.claimFee();
+        Token18 token = market.token();
+        token.push(comptroller, token.balanceOf());
```
kbrizzle, line 41: nit: `token.push(comptroller);` will push the whole balance

kbrizzle: you can inline the `market.token().push(comptroller)` then too

PR #253, opened 2024-02-16
Add subtractive fees

kbrizzle:


packages/perennial/contracts/MarketFactory.sol
```diff
@@ -57,6 +60,14 @@ contract MarketFactory is IMarketFactory, Factory {
         emit OperatorUpdated(msg.sender, operator, newEnabled);
     }
 
+    /// @notice Updates the status of an operator for the caller
```
arjun-io, line 63: nit: fix natspec here (using operator description, unfinished param below)

packages/perennial/contracts/Market.sol
```diff
@@ -408,7 +450,13 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         context.local.update(collateral);
 
         // protect account
-        if (newOrder.protected()) liquidators[account][context.local.currentId] = msg.sender;
+        if (newOrder.protected()) updateContext.liquidator = msg.sender;
+
+        // apply referrer
+        if (updateContext.referrer == address(0))
```
arjun-io, line 456: I think this is DoS'able if any msg.sender sends an update with a tiny amount of collateral and a referral target, since we don't check if the subtractive fee is non-zero

packages/perennial/contracts/Market.sol
```diff
@@ -476,6 +471,37 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
+    /// @param updateContext The update context to use
+    /// @param newOrder The order to process
+    /// @param referrer The referrer of the order
+    function _processReferrer(
+        UpdateContext memory updateContext,
+        Order memory newOrder,
+        address referrer
+    ) private pure {
+        if (newOrder.makerReferral.isZero() && newOrder.takerReferral.isZero()) return;
+        if (updateContext.referrer == address(0)) { updateContext.referrer = referrer; return; }
```
arjun-io, line 499: nit: I think you could omit this `return` to keep the `if` a single line without brackets since the next line will handle the early return after the updateContext is modified, that might be a bit more complex to understand though

arjun-io (approved):

arjun-io (approved):

packages/perennial/contracts/Market.sol
kbrizzle: good call 👍

PR #278, opened 2024-03-21
Market tests for exposure when maker is zero

EdNoepel:
A market incurs exposure when it enters a state in which taker imbalance is not covered by maker liquidity.  This state cannot be archieved by opening/closing positions; an undercollateralized maker must be liquidated to arrive at this state.  This change extends the global assertion wrapper to cover `exposure`, and adds two tests:
- Updating risk parameters to add adiabatic fees after a maker is liquidated
- Liquidating a maker in a market where adiabatic fees have already been established

### DIFF
--- a/packages/perennial/test/integration/core/fees.test.ts
+++ b/packages/perennial/test/integration/core/fees.test.ts
@@ -147,6 +147,7 @@ describe('Fees', () => {
         riskFee: expectedRiskFee,
         oracleFee: expectedOracleFee,
         donation: expectedDonation,
+        exposure: 0,
       })
       expectOrderEq(await market.pendingOrder(2), {
         ...DEFAULT_ORDER,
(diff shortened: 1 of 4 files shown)


EdNoepel:
Discovered I have many more `expectGlobalEq` calls to update.  Hope to complete this afternoon.

packages/perennial/test/integration/core/fees.test.ts
```diff
@@ -1016,6 +1023,7 @@ describe('Fees', () => {
         riskFee: expectedRiskFee,
         oracleFee: expectedOracleFee,
         donation: expectedDonation,
+        exposure: 0,
```
kbrizzle, line 1026: Could you make a follow up ticket to create a `DEFAULT_GLOBAL` so that we don't have to explicitly define all these `0` value fields everywhere?

kbrizzle (approved):

PR #279, opened 2024-03-22
Unit test to liquidate user under minMaintenance

EdNoepel:
`minMaintenance` is a fixed-amount risk parameter.   To ensure liquidations are incentivized, user must maintain at least that amount of collateral regardless of the maintenance requirements for their position.  Implemented a unit test which drops a user's collateral under the `minMaintenance` amount and then liquidates the user.

### DIFF
--- a/packages/perennial/test/integration/core/liquidate.test.ts
+++ b/packages/perennial/test/integration/core/liquidate.test.ts
@@ -3,10 +3,11 @@ import 'hardhat'
 import { BigNumber, constants, utils } from 'ethers'
 
 import { InstanceVars, deployProtocol, createMarket, settle } from '../helpers/setupHelpers'
-import { parse6decimal } from '../../../../common/testutil/types'
+import { expectPositionEq, parse6decimal } from '../../../../common/testutil/types'
 import { loadFixture } from '@nomicfoundation/hardhat-network-helpers'
 
 export const TIMESTAMP_2 = 1631113819
+export const TIMESTAMP_3 = 1631114005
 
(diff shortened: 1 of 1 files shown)


packages/perennial/test/integration/core/liquidate.test.ts
```diff
@@ -222,4 +223,71 @@ describe('Liquidate', () => {
...
+    await chainlink.nextWithPriceModification(price => price.div(2))
+    await settle(market, user)
+    await chainlink.nextWithPriceModification(price => price.div(4))
+    // TODO: replace setupHelper.settle with this implementation, renaming the existing
+    await market.connect(user)['settle(address)'](user.address) // avoid invariant violation
+
+    // ensure user's collateral is now lower than minMaintenance
+    const collateral = (await market.locals(user.address)).collateral
+    const riskParameter = await market.riskParameter()
+    expect(collateral).to.be.lessThan(riskParameter.minMaintenance)
```
kbrizzle, line 268: This case may already work for this, but could you ensure that both:
- the collateral is under `minMaintenance`.
- the collateral would otherwise not be under `maintenance` if calculated traditionally without a minimum.

If this case doesn't work, it's probably easiest to create a small position that's notional is about the same size as `minMaintenance` then lose a little bit of collateral.

EdNoepel: Done in ca67e2fb1e3ad77a0f326315f5578a9e827ad797.

kbrizzle (approved):

PR #280, opened 2024-03-26
Integration tests for referral fee

EdNoepel:
Tests ensure the proper amount of the base fee is claimable by the referrer after account settlement.

### DIFF
--- a/packages/perennial/test/integration/core/fees.test.ts
+++ b/packages/perennial/test/integration/core/fees.test.ts
@@ -1696,4 +1696,139 @@ describe('Fees', () => {
       ).to.equal(expectedFunding)
     })
   })
+
+  describe('referral fees', () => {
+    const COLLATERAL = parse6decimal('600')
+    const POSITION = parse6decimal('3')
+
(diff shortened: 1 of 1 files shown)


packages/perennial/test/integration/core/fees.test.ts
```diff
@@ -1696,4 +1696,179 @@ describe('Fees', () => {
       ).to.equal(expectedFunding)
     })
   })
+
+  describe('referral fees', () => {
```
kbrizzle, line 1700: Could you add a case or two testing two different referred orders coming in at the same version? even better if the two orders have different referral fee rates.

Would be good to make sure we're aggregating / splitting these properly all the way through.

kbrizzle (approved):

packages/perennial/test/integration/core/fees.test.ts
EdNoepel: Done in 420bd5025a653e242112c1995dee7d88229df271

PR #283, opened 2024-03-27
Update settle unit test helper

EdNoepel:
**Changes**
- Update existing settle helper to actually call `settle`, which was new for 2.2.
- Introduce new `updateNoOp` helper replacing the old call.  Use it in places where changes to position IDs would be confusing.
- Fix issue with interest and funding fee tests, which were only looking at the first emitted event of the desired type.  Use a map-reduce solution to parse and aggregate desired fields from events.

### DIFF
--- a/packages/perennial/test/integration/core/fees.test.ts
+++ b/packages/perennial/test/integration/core/fees.test.ts
@@ -796,7 +796,7 @@ describe('Fees', () => {
 
       // Settle maker to give them portion of fees
       await nextWithConstantPrice()
-      await settle(market, user)
+      await updateNoOp(market, user)
 
       await expect(
         market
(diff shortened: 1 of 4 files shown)


packages/perennial/test/integration/core/fees.test.ts
```diff
@@ -1575,24 +1596,31 @@ describe('Fees', () => {
...
+        BigNumber.from(0),
+      )
+      expect(accumulatedInterest).to.equal(expectedInterest.mul(-1))
+      expect(accumulatedInterestFee).to.equal(expectedInterestFee)
+      expect(accumulatedInterestMaker).to.equal(expectedInterest)
     })
   })
 
-  describe('funding fee', () => {
+  describe.only('funding fee', () => {
```
kbrizzle, line 1623: nit: .only

EdNoepel: My bad; fixed in 17faf3d593e7db89554330578b2ad4d05e97adb3.

EdNoepel:
Cannot reproduce the oracle package CI error; suspect it's an intermittent JSON-RPC issue with the integration test.

kbrizzle (approved):

kbrizzle:
Yeah there are a few flaky tests throughout, so feel free to rerun failed CI jobs as needed if something feels off.

PR #284, opened 2024-03-27
Remove v2.2 migration code

EdNoepel:
## Changes
- Removed slot 1 from `Local` storage
- Eliminate loading of collateral data from `Local` into new `Checkpoint`
- Eliminate `Market::updateOracle()`

### Local
Slot 0 unchanged.  Slot 1 was deprecated in 2.2, where it was updated to write 0's.  Slot 1 is now eliminated.
Facility introduced in 2.2 to initialize `Checkpoint` with collateral from `Local` has been removed.  **Must go into no-order mode and settle all users prior to deployment.**

### DIFF
--- a/packages/perennial/contracts/Market.sol
+++ b/packages/perennial/contracts/Market.sol
@@ -144,14 +144,6 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         _storeContext(context, account);
     }
 
-    /// @notice Updates the oracle of the market
-    /// @dev For the v2.1.1 -> v2.2 migration process, not to be used otherwise
-    /// @param newOracle The new oracle address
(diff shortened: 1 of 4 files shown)


kbrizzle (approved):

kbrizzle:
Instead of waiting to merge into the `v2.3` branch until all state has been settled up, let's add a migration note similar to this PR: https://github.com/equilibria-xyz/perennial-v2/pull/208 with all information we might need to know about what's changing and what is required to ensure a safe migration.

We'll scan back through these notes and create a `v2.3` migration runbook like [this](https://github.com/equilibria-xyz/perennial-v2/blob/v2.2/runbooks/MIGRATION_v2.2.md) when we're ready to go to audit.

PR #287, opened 2024-03-29
Remove maker/takerCloseAlways

EdNoepel:
Remove two market parameter knobs, adjusting logic such that makerCloseAlways=false and takerCloseAlways=true.  Reworked unit tests, removing those which specifically tested the behavior of these knobs.

### DIFF
--- a/packages/perennial/contracts/types/Order.sol
+++ b/packages/perennial/contracts/types/Order.sol
@@ -167,11 +167,21 @@ library OrderLib {
     /// @return Whether the order is applicable for liquidity checks
     function liquidityCheckApplicable(
         Order memory self,
-        MarketParameter memory marketParameter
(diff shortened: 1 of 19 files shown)


packages/perennial/contracts/types/Order.sol
```diff
@@ -167,11 +167,14 @@ library OrderLib {
     /// @return Whether the order is applicable for liquidity checks
     function liquidityCheckApplicable(
         Order memory self,
-        MarketParameter memory marketParameter
+        MarketParameter memory marketParameter // TODO: consider changing this to a isMarketClosed bool to save gas
```
kbrizzle, line 170: nit: TODO

EdNoepel: removed in 0b45707ba1c03f25d26e9d8e0efdbb51cdb0545e

kbrizzle (approved):

PR #288, opened 2024-03-29
Collapse Context and UpdateContext structs

EdNoepel:
For code size reduction.

### DIFF
--- a/packages/perennial/contracts/Market.sol
+++ b/packages/perennial/contracts/Market.sol
@@ -320,12 +320,12 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         context.local = _locals[account].read();
 
         // latest positions
-        context.latestGlobalPosition = _position.read();
-        context.latestLocalPosition = _positions[account].read();
+        context.latestPositionGlobal = _position.read();
+        context.latestPositionLocal = _positions[account].read();
 
         // aggregate pending orders
-        context.pendingGlobalOrder = _pending.read();
-        context.pendingLocalOrder = _pendings[account].read();
+        context.pendingGlobal = _pending.read();
+        context.pendingLocal = _pendings[account].read();
     }
 
     /// @notice Stores the context for the transaction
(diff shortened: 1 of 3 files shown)


EdNoepel:
Ran all the integration tests locally; all pass.  Have rerun some of these CI jobs several times, but keep having `HttpProviderError` failures.  Recommend checking Alchemy dashboard to investigate.

packages/perennial/contracts/Market.sol
```diff
@@ -409,42 +409,42 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
             collateral,
             newMaker,
             newLong,
             newShort,
             protect,
             referralFee
         );
-        updateContext.currentPosition.global.update(newOrder);
-        updateContext.currentPosition.local.update(newOrder);
+        updateContext.currentGlobalPosition.update(newOrder);
```
kbrizzle, line 440: Could we keep the naming consistent for now: `currentPosition.local` -> `currentPositionLocal`.

packages/perennial/contracts/Market.sol
```diff
@@ -409,42 +409,42 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
+        updateContext.currentLocalPosition.update(newOrder);
 
         // apply new order
-        updateContext.order.local.add(newOrder);
-        updateContext.order.global.add(newOrder);
-        context.pending.global.add(newOrder);
-        context.pending.local.add(newOrder);
+        updateContext.orderLocal.add(newOrder);
+        updateContext.orderGlobal.add(newOrder);
+        context.pendingGlobalOrder.add(newOrder);
```
kbrizzle, line 446: Here as well `pending.global` -> `pendingGlobal`.

packages/perennial/contracts/Market.sol
```diff
@@ -513,9 +513,9 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         address account
     ) private view returns (SettlementContext memory settlementContext) {
         // processing accumulators
-        settlementContext.latestVersion = _versions[context.latestPosition.global.timestamp].read();
-        settlementContext.latestCheckpoint = _checkpoints[account][context.latestPosition.local.timestamp].read();
-        settlementContext.orderOracleVersion = oracle.at(context.latestPosition.global.timestamp);
+        settlementContext.latestVersion = _versions[context.latestGlobalPosition.timestamp].read();
```
kbrizzle, line 516: And here `latestPosition.global` -> `latestPositionGlobal`.

kbrizzle:
Left some nits on naming, I'm thinking we should keep things as is for now. We may want to readdress these after we complete the `currentPosition` cleanup task.

The intuition here is that there are three things we're keeping track of:
- `position` -> the settled position.
- `order` -> the current order we're operating on.
- `pending` -> the aggregate value of all of the pending orders.

`currentPosition` is the odd one out here that's not really something that should be loaded in the Context, but is for now for cleanliness since it's used throughout. Since we need this we differentiate between `position` and `currentPosition` with a `latest` prefix.

kbrizzle (approved):

PR #292, opened 2024-04-09
[Hotfix] Disable deposits and redemptions in same oracle version

arjun-io:


packages/perennial-vault/test/unit/types/Checkpoint.test.ts
```diff
@@ -241,7 +241,7 @@ describe('Checkpoint', () => {
   })
 
   describe('#update', () => {
-    it('updates the checkpoint', async () => {
+    it.skip('updates the checkpoint', async () => {
```
kbrizzle, line 244: can we make two more cases here to cover redeem only and deposit only, just so we still have coverage

kbrizzle (approved):

PR #296, opened 2024-04-11
[v2.2 Audit Fix] KeeperOracle flag to trap invalid oracle versions

EdNoepel:
## Changes
- Establish unit tests for covering `KeeperOracle` without focusing on a specific (Pyth/Chainlink) implementation.
- Reproduce [audit issue](https://github.com/sherlock-audit/2024-02-perennial-v2-3-judging/issues/6) handling of expired prices.  Instead of mapping validity as suggested by the reporter, this implementation maps invalidity.  This reduces gas costs for keepers persisting non-expired values.
- Clean up redundant initialization logic in `PythOracleFactory.test.ts` (unrelated).

## Deployment Plan
`KeeperFactory` and `KeeperOracle` and  instances will need to be redeployed.  Suspect this was already part of the 2.1 -> 2.2 migration plan.

### DIFF
--- a/packages/perennial/contracts/Market.sol
+++ b/packages/perennial/contracts/Market.sol
@@ -543,7 +543,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
 
         Order memory nextOrder;
 
-        // settle
+        // settle - process orders whose requested prices are now available from oracle
         while (
             context.global.currentId != context.global.latestId &&
             (nextOrder = _pendingOrder[context.global.latestId + 1].read()).ready(context.latestOracleVersion)
@@ -554,7 +554,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
             (nextOrder = _pendingOrders[account][context.local.latestId + 1].read()).ready(context.latestOracleVersion)
         ) _processOrderLocal(context, settlementContext, account, context.local.latestId + 1, nextOrder);
 
-        // sync
+        // sync - advance position timestamps with the latest oracle version
         if (context.latestOracleVersion.timestamp > context.latestPosition.global.timestamp) {
             nextOrder = _pendingOrder[context.global.latestId].read();
             nextOrder.next(context.latestOracleVersion.timestamp);


EdNoepel:
For some reason, running oracle units tests with coverage results in gas price errors, but running without coverage works fine.  Can reproduce locally; looking into it.

packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
```diff
@@ -26,6 +26,9 @@ contract KeeperOracle is IKeeperOracle, Instance {
     /// @dev Mapping from oracle version to oracle version data
     mapping(uint256 => Fixed6) private _prices;
 
+    /// @dev Flags expired versions as invalid, indicating they hold a carryover price
+    mapping(uint256 => bool) private _invalidVersions;
```
arjun-io, line 30: Can we combine these two mappings (prices and invalid versions)? I believe this would be safe since these implementations haven't been deployed yet 

Potentially we store both `price` and `valid`

packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
```diff
@@ -108,7 +111,7 @@ contract KeeperOracle is IKeeperOracle, Instance {
     /// @return oracleVersion Oracle version at version `version`
     function at(uint256 timestamp) public view returns (OracleVersion memory oracleVersion) {
         (oracleVersion.timestamp, oracleVersion.price) = (timestamp, _prices[timestamp]);
-        oracleVersion.valid = !oracleVersion.price.isZero();
+        oracleVersion.valid = !oracleVersion.price.isZero() && !_invalidVersions[timestamp];
```
arjun-io, line 114: Should we only use the invalid version flag here?

packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
EdNoepel: I did this in branch `ed/PE-1303-struct`, but do not recommend.
- Because `Fixed6` takes up an entire storage slot, each value in the mapping will consume two slots.  To fit both in a single slot, we could use `i248` and `wrap`/`unwrap` everywhere, but that would introduce other concerns.
- With this implementation, we're forcing keepers to consistently spend extra gas upon each 'commit', rather than just spending a bit extra on the corner case.

Here's a gas comparison:
# Separate invalidity map (this branch)
·---------------------------------------|----------------------------|-----------------|-----------------------------·
|         Solc version: 0.8.24          ·  Optimizer enabled: false  ·  Runs: 1000000  ·  Block limit: 30000000 gas  │
········································|····························|·················|······························
|  Methods                                                                                                           │
·····················|··················|··············|·············|·················|···············|··············
|  Contract          ·  Method          ·  Min         ·  Max        ·  Avg            ·  # calls      ·  usd (avg)  │
·····················|··················|··············|·············|·················|···············|··············
|  ChainlinkFactory  ·  commit          ·           -  ·          -  ·         179325  ·            3  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  KeeperOracle      ·  commit          ·       83511  ·      98477  ·          88500  ·            6  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  Oracle            ·  request         ·       41672  ·     241605  ·          78376  ·           23  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  PythFactory       ·  commit          ·      151444  ·     151540  ·         151492  ·            2  ·          -  │
·····················|··················|··············|·············|·················|···············|··············

# Single map (ed/PE-1303-struct)
·---------------------------------------|----------------------------|-----------------|-----------------------------·
|         Solc version: 0.8.24          ·  Optimizer enabled: false  ·  Runs: 1000000  ·  Block limit: 30000000 gas  │
········································|····························|·················|······························
|  Methods                                                                                                           │
·····················|··················|··············|·············|·················|···············|··············
|  Contract          ·  Method          ·  Min         ·  Max        ·  Avg            ·  # calls      ·  usd (avg)  │
·····················|··················|··············|·············|·················|···············|··············
|  ChainlinkFactory  ·  commit          ·           -  ·          -  ·         201515  ·            3  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  KeeperOracle      ·  commit          ·       78521  ·     105700  ·          96636  ·            6  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  Oracle            ·  request         ·       41672  ·     241605  ·          78376  ·           23  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  PythFactory       ·  commit          ·      173634  ·     173730  ·         173682  ·            2  ·          -  │
·····················|··················|··············|·············|·················|···············|··············

packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
EdNoepel: We currently allow keepers to commit a "valid" 0 price, which `at` would mark invalid.  Happy to remove the check here if we'd like "real" 0 prices coming from the oracle to be accepted by the market.

EdNoepel: I tried this in a local branch by initializing timestamp 0 as invalid.  Even with that, we run into trouble with `returns invalid version if that version was not requested` and `returns invalid version if that version was requested but not committed` tests.  These four tests expect these non-existant prices to be invalid, which seems like the appropriate behavior.

Resolving this would either involve writing additional state upon every commit, either by storing a validity flag alongside the price (as discussed above), or by switching from an invalidity map to a validity map (which seems less desirable).

packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
kbrizzle: We're going to want to pack the price / valid here I'm pretty sure.

`price` should only need 64 bits, so there's plenty of room for both in one storage slot -- we use standard full-slot number types in memory because there's no packing for in-memory data.

I did a little bit of scratch work [here](https://github.com/equilibria-xyz/perennial-v2/commit/7400820aec425037132227f51f1aa6f834ca6ed8) and this saved 20k+ gas over your struct implementation by introducing a `Price` type.

EdNoepel: Confirmed; gas cost with your `PriceStorageLib` implementation is similar to this PR:
```
·---------------------------------------|----------------------------|-----------------|-----------------------------·
|         Solc version: 0.8.24          ·  Optimizer enabled: false  ·  Runs: 1000000  ·  Block limit: 30000000 gas  │
········································|····························|·················|······························
|  Methods                                                                                                           │
·····················|··················|··············|·············|·················|···············|··············
|  Contract          ·  Method          ·  Min         ·  Max        ·  Avg            ·  # calls      ·  usd (avg)  │
·····················|··················|··············|·············|·················|···············|··············
|  ChainlinkFactory  ·  commit          ·           -  ·          -  ·         180390  ·            3  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  KeeperOracle      ·  commit          ·       77862  ·      85031  ·          82641  ·            6  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  Oracle            ·  request         ·       41672  ·     241605  ·          78375  ·           23  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
|  PythFactory       ·  commit          ·      152509  ·     152605  ·         152557  ·            2  ·          -  │
·····················|··················|··············|·············|·················|···············|··············
```

I'll close this and continue work in the other branch.

PR #303, opened 2024-04-14
[v2.2 Audit Fix] Resolves double-subtraction issue with asset ineligibility calculation

kbrizzle:
Resolves: https://github.com/sherlock-audit/2024-02-perennial-v2-3-judging/issues/11.

Adds additional documentation on what is going on in this calculation and why we do not need to the second `.add(withdrawal)`.


arjun-io (approved):

EdNoepel (approved):
Was a problem prior to this PR, but perhaps we should consider correcting spelling to _ineligible_ and _eligible_?

kbrizzle:
oops, yeah went through and fixed the spelling.

PR #305, opened 2024-04-14
[v2.2 Audit Fix] Split global settlement fee between deposits and redemptions

kbrizzle:
Splits the order count tracking in the Vault checkpoints into separate deposit and redeem counts. This allows us to split the global settlement fee for the id properly between deposits and redemptions.

Resolves https://github.com/sherlock-audit/2024-02-perennial-v2-3-judging/issues/26.

### Migration Note
- Splits `Checkpoint.orders` into `Checkpoint.deposits` and `Checkpoint.redemptions`
  - The settle-only migration step will ensure that all legacy checkpoints have been read before switching to the new format

### DIFF
--- a/packages/perennial-vault/contracts/types/Checkpoint.sol
+++ b/packages/perennial-vault/contracts/types/Checkpoint.sol
@@ -98,9 +98,10 @@ library CheckpointLib {
         if (self.shares.isZero()) return assets;
 
         // if vault is insolvent, default to par value
+        UFixed6 settlementFee = _settlementFeeForOrders(self, self.deposits);
         return self.assets.lte(Fixed6Lib.ZERO) ?
-            assets.unsafeSub(self.settlementFee) :
-            _toShares(self, assets).unsafeSub(_toSharesExact(self, self.settlementFee));
+            assets.unsafeSub(settlementFee) :
+            _toShares(self, assets).unsafeSub(_toSharesExact(self, settlementFee));
     }
 
     /// @notice Converts a given amount of shares to assets with checkpoint in the global context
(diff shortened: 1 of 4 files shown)


arjun-io (approved):

EdNoepel (approved):
General question regarding migration, which could be applied to most storage changes, not just this.

When market goes into _settle only_ mode, assume we manually settle all accounts, storing _orders_ in the `Checkpoint`.  We then upgrade contracts.  Now all new `Checkpoint`s will store deposits were orders once were, and have the new _redemptions_ field.

If someone calls `Market::checkpoints`, is it okay that orders will now be reported as deposits?  How will subgraph indexer handle this change?

arjun-io:
> General question regarding migration, which could be applied to most storage changes, not just this.
> 
> When market goes into _settle only_ mode, assume we manually settle all accounts, storing _orders_ in the `Checkpoint`. We then upgrade contracts. Now all new `Checkpoint`s will store deposits were orders once were, and have the new _redemptions_ field.
> 
> If someone calls `Market::checkpoints`, is it okay that orders will now be reported as deposits? How will subgraph indexer handle this change?

Yeah this bit of context will need to be shared with integrators to see if/how they should handle it - especially those who do smart contract integration. On the indexing side this isn't as big of a deal - we will either [graft](https://thegraph.com/docs/en/cookbook/grafting/) the new subgraph onto the old one or treat indexing as a series of "hardforks" of implementations with block ranges corresponding to different Perennial versions - the `eth_call`s are performed with the block number param so we don't need to handle things differently there

PR #307, opened 2024-04-16
Add intent order fills

kbrizzle:
- Adds `perennial-verifier` package which produces a `Verifier` contract to be used for ERC712 payload verification
- Adds `update()` overload that takes a signed intent and fills both the marker and taker orders
- Renames `Intent` -> `Guarantee` to avoid name collision with signed payloads

### TODO:
- [x] Add test suite for Verifier
- [x] Add tests for new update fill method in Market
- [x] Add tests for guarentees in Market


packages/perennial-verifier/contracts/Verifier.sol
```diff
@@ -0,0 +1,96 @@
...
+import { IVerifier } from "./interfaces/IVerifier.sol";
+
+// TODO: permit cancel nonce / group
+// TODO: permit operator approval
+
+/// @title Verifier
+/// @notice Singleton ERC712 signed message verifier for the Perennial protocol.
+/// @dev Handles nonce management for verified messages.
+///       - nonce is a single use unique value per message that is invalidated after use
+///       - group is allows for an entire set of messages to be invalidated via a single cancel operation
```
arjun-io, line 18: nit: typo `group allows` (remove `is`)

packages/perennial-verifier/.github/workflows/test.yml
arjun-io, line 1: Is this copied from somewhere else or do we want to use foundry for this package's CI?

packages/perennial-verifier/contracts/Verifier.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @notice Cancels a group nonce
+    /// @param account The account to cancel the group nonce for
+    /// @param group The group nonce to cancel
+    function _cancelGroup(address account, uint256 group) private {
+        groups[account][group] = true;
+        emit GroupCancelled(account, group);
+    }
+
+    /// @dev Validates the common data of a message
+    modifier validate(Common calldata common, bytes calldata signature) {
```
arjun-io, line 85: nit: maybe `validateAndCancel` to note that there's a side effect in this function which cancels the nonce?

packages/perennial-verifier/contracts/interfaces/IVerifier.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+
+import { Intent } from "../types/Intent.sol";
+import { Fill } from "../types/Fill.sol";
+
+interface IVerifier {
+    error VerifierInvalidDomainError();
```
arjun-io, line 8:
```suggestion
    // sig: 0xb09262f6
    error VerifierInvalidDomainError();
```

packages/perennial-verifier/contracts/interfaces/IVerifier.sol
```diff
@@ -0,0 +1,21 @@
...
+    error VerifierInvalidSignatureError();
```
arjun-io, line 9:
```suggestion
    // sig: 0xb09262f6
    error VerifierInvalidSignatureError();
```

packages/perennial-verifier/contracts/interfaces/IVerifier.sol
```diff
@@ -0,0 +1,21 @@
...
+    error VerifierInvalidNonceError();
```
arjun-io, line 10:
```suggestion
    // sig: 0xe6784f14
    error VerifierInvalidNonceError();
```

packages/perennial-verifier/contracts/interfaces/IVerifier.sol
```diff
@@ -0,0 +1,21 @@
...
+    error VerifierInvalidGroupError();
```
arjun-io, line 11:
```suggestion
    // sig: 0x79998279
    error VerifierInvalidGroupError();
```

packages/perennial-verifier/contracts/interfaces/IVerifier.sol
```diff
@@ -0,0 +1,21 @@
...
+    error VerifierInvalidExpiryError();
```
arjun-io, line 12:
```suggestion
    // sig: 0x27661908
    error VerifierInvalidExpiryError();
```

packages/perennial-verifier/README.md
```diff
@@ -0,0 +1,62 @@
+# Perennial V2 Extensions
```
arjun-io, line 1:
```suggestion
# Perennial V2 Verifier
```

packages/perennial-verifier/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Extension systems for the Perennial V2 Protocol
```
arjun-io, line 3:
```suggestion
Intent and Fill Message verifier contracts for the Perennial V2 Protocol
```

packages/perennial/contracts/Market.sol
```diff
@@ -107,6 +116,55 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         _storeContext(context, account);
     }
 
+    /// @notice Updates the maker and taker positions of an intent order
```
arjun-io, line 119: `maker` might be a bit of an overloaded term here, since we're actually just opening an opposite side taker position here rather than a maker position

packages/perennial/contracts/Market.sol
```diff
@@ -107,6 +116,55 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
+    /// @param intent The intent that is being filled
+    /// @param signature The signature of the intent that is being filled
+    /// @param referrer The referrer of the order
```
arjun-io, line 122: It seems like the referrer should be set by the intent account rather than the filler?

packages/perennial-verifier/contracts/Verifier.sol
kbrizzle: Added a follow up task to remove this from extensions as well

packages/perennial-verifier/.github/workflows/test.yml
kbrizzle: Added a follow up task to remove this from extensions as well

packages/perennial/contracts/Market.sol
kbrizzle: Ok yeah changing this to `sender` / `signer` instead. There is a difference in that the "maker" or "sender" role here pays the fees, while the "signer" or "taker" does not (instead having it baked into their quote).

packages/perennial/contracts/Market.sol
kbrizzle: Good point, I'm going to leave this one as is for now, but I included in my follow up task for adding a solver network fee to figure this out as well.

PR #308, opened 2024-04-16
[v2.2 Audit Fix] KeeperOracle flag to trap invalid oracle versions

EdNoepel:
## Changes
- Establish unit tests for covering `KeeperOracle` without focusing on a specific (Pyth/Chainlink) implementation.
- Reproduce [audit issue](https://github.com/sherlock-audit/2024-02-perennial-v2-3-judging/issues/6) handling of expired prices.
- Resolve by storing a validity flag alongside price.
- Clean up redundant initialization logic in `PythOracleFactory.test.ts` (unrelated).

## Deployment Plan
`KeeperFactory` and `KeeperOracle` and  instances will need to be redeployed.  Suspect this was already part of the 2.1 -> 2.2 migration plan.

### DIFF
--- a/packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
+++ b/packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
@@ -169,8 +169,6 @@ contract KeeperOracle is IKeeperOracle, Instance {
         if (!version.valid) revert KeeperOracleInvalidPriceError();
         if (version.timestamp <= _global.latestVersion || (next() != 0 && version.timestamp >= next()))
             revert KeeperOracleVersionOutsideRangeError();
-        if (version.price.gt(Fixed6.wrap(type(int64).max)) || version.price.lt(Fixed6.wrap(type(int64).min)))
-            revert KeeperOracleInvalidPriceError();
 
         _prices[version.timestamp].store(Price(version.price, true));
         return false;
--- a/packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
+++ b/packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
@@ -14,6 +14,15 @@ import "../interfaces/IKeeperFactory.sol";
 contract KeeperOracle is IKeeperOracle, Instance {
     using EnumerableSet for EnumerableSet.AddressSet;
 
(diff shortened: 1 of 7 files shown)


packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
```diff
@@ -168,7 +169,10 @@ contract KeeperOracle is IKeeperOracle, Instance {
         if (!version.valid) revert KeeperOracleInvalidPriceError();
         if (version.timestamp <= _global.latestVersion || (next() != 0 && version.timestamp >= next()))
             revert KeeperOracleVersionOutsideRangeError();
-        _prices[version.timestamp] = version.price;
+        if (version.price.gt(Fixed6.wrap(type(int64).max)) || version.price.lt(Fixed6.wrap(type(int64).min)))
```
arjun-io, line 172: Can we rely on the `store` errors here to throw?

packages/perennial-oracle/test/unit/oracle/KeeperOracle.test.ts
arjun-io, line 1: 🔥

packages/perennial-oracle/contracts/keeper/KeeperOracle.sol
EdNoepel: Yes; I am in favor of removing line 172.  Loosely related, I need to update unit tests to cover negative prices.

EdNoepel: done in 7f2d3bdb45975e428b64df0899158bba1105fa51

arjun-io (approved):

kbrizzle (approved):

kbrizzle:
Just to close the loop here, @arjun-io and I thought of two additional reasons why we need to use this packed-struct version of the price storage over the dual-storage method:

- While it's true that the "normal case" would be slightly cheaper in the dual model, we would like each call to commit to cost roughly the same amount so that it's easier to normalize the keeper incentivization payouts.
- `at()` calls would be more expensive in the dual case since we need to read both slots, and this is the more sensitive flow for cost reduction.

PR #309, opened 2024-04-17
[v2.2 Audit Fix] Oracle and Vault to use settle instead of no-op update

EdNoepel:
Merged a 2.3 PR which updates `setupHelpers::settle` to actually call `Market::settle` rather than a no-op update.

Updates `KeeperOracle::_settle`, `Vault::rebalance`, and `Vault::update` to call `Market::settle` instead of performing a no-op update.  Resolves https://github.com/sherlock-audit/2024-02-perennial-v2-3-judging/issues/30 and https://github.com/sherlock-audit/2024-02-perennial-v2-3-judging/issues/23.

Explicitly did _not_ update `InvariantLib` to exempt empty orders from the margin check, as we wish to dissuade integrators from using no-op orders for settlement.

Made some changes to vault liquidation tests:
- Renamed tests and adjusted comments to clarify the direction of the taker position and exposure of the maker position.
- Removed `Vault::update` statements which comments explained as "We should no longer be able to deposit or redeem", but were not actually testing either.  Replaced them with a new test which assured unexposed positions _can_ be reedeemed regardless of margin.


packages/perennial-vault/test/integration/vault/Vault.test.ts
```diff
@@ -691,6 +691,20 @@ describe('Vault', () => {
...
+      // Deposit 100k into the vault and establish initial oracle price (38838.362695).
+      const depositAmount = parse6decimal('100000')
+      await vault.connect(user).update(user.address, depositAmount, 0, 0)
+      await updateOracle()
+
+      // Increase the oracle price such that the vault's maker position would violate margin requirements
+      // if it settled and had short exposure.
+      await updateOracle(undefined, parse6decimal('50000'))
+
+      // Since the position was never settled, ensure user can withdraw from vault.
```
kbrizzle, line 705: This might need some deeper debugging, after thinking through this more, I don't think this is correct.

Original thought was:
- (1) The vault will only settle() the underlying markets, which always works
- (2) The proceeding update() is to fully close each position in the underlying markets
  - Caveat here is that you likely have accrued some dust on top of the `depositAmount`, however it's small enough that you're under the minMargin threshold and therefore will have a new pending position of `zero`.
- (3) A full closure is allowed as an update when liquidatable

After looking through the invariant again, [(3) does not look to be true](https://github.com/equilibria-xyz/perennial-v2/blob/v2.2/packages/perennial/contracts/libs/InvariantLib.sol#L80). So without passing in a protected flag (which we shouldn't), I don't see how that's not reverting. Either I'm missing something, or maybe the position is not actually becoming  liquidatable in our setup.

packages/perennial-vault/test/integration/vault/Vault.test.ts
```diff
@@ -1910,18 +1924,16 @@ describe('Vault', () => {
...
+      context('short exposure', () => {
         it('recovers from a liquidation', async () => {
           await vault.connect(user).update(user.address, parse6decimal('100000'), 0, 0)
           await updateOracle()
 
-          // 1. An oracle update makes the long position liquidatable.
-          // We should now longer be able to deposit or redeem
+          // 1. An oracle update increases the price, making the vault's position liquidatable.
           await updateOracle(undefined, parse6decimal('50000'))
-          await expect(vault.connect(user).update(user.address, 0, 0, 0)).to.be.reverted
```
kbrizzle, line 1921: This is no longer reverting because:
- flywheel no longer calls update() -> settle()
- claiming no longer triggers a position rebalance (as of v2.2) -> a no-op update in the vault is registered as a claim.

It might be worth checking that:
- `vault.connect(user).update(user.address, 0, 0, 0)` now does not revert
- `vault.connect(user).update(user.address, 2, 0, 0)` would revert still

packages/perennial-vault/test/integration/vault/Vault.test.ts
```diff
@@ -1955,17 +1963,14 @@ describe('Vault', () => {
           expect(await btcCollateralInVault()).to.equal(btcFinalCollateral)
         })
 
-        it('recovers from a liquidation w/ shortfall', async () => {
+        it('recovers from a liquidation w/ shortfalll', async () => {
```
kbrizzle, line 1966: typo: shortfalll

packages/perennial-vault/test/integration/vault/Vault.test.ts
EdNoepel: The latter (deposit) does revert, but it's due to insufficient funds, not a margin issue.  When I called `fundWallet` on the address beforehand, I _was_ able to deposit in this use case.

packages/perennial-vault/test/integration/vault/Vault.test.ts
EdNoepel: @kbrizzle  There are a few prices at play here:
38k - vault's position is healthy
50k - vault is still within _margin_ requirements (users may deposit/redeem), yet under _maintenance_ requirements (making it liquidatable)
99k - vault is under both _margin_ and _maintenance_ requirements, so users may not deposit/reedeem, and it remains liquidatable

I would like to change this price to 99k and rework this test to exercise `StrategyLibInsufficientCollateralError`, which isn't currently covered.
As for the liquidation tests, I would like to keep them at 50k and omit the deposit/withdrawal testing.
Unless, of course, the behavior I described above is unexpected.

kbrizzle: For the 50k level - the margin requirement should be higher than the maintenance requirement, so that would get violated first not second between the two. Basically you want to cut off the ability to increase leverage a bit before the actual liquidation point for UX.

packages/perennial-vault/test/integration/vault/Vault.test.ts
kbrizzle: oops you're right, a trivial redeem should work here instead without having to re-fund the wallet:

`vault.connect(user).update(user.address, 0, 2, 0)`

packages/perennial-vault/test/integration/vault/Vault.test.ts
EdNoepel: Both margin and maintenance ratios are set to 0.10, so that wasn't the problem.

It is impossible to attempt a redeem without implicitly rebalancing.  Here, the implicit rebalance reduces the position and adds collateral.  When the redeem invariant check occurs, collateral has been deposited, the position has not yet been reduced, but the vault is well above margin requirements.

```
updating oracle price to 50k
testing small redeem
  PositionLib calculating collateral reqd for position 2.059819 * price 50000.000000 * ratio .100000
Vault targeting 1.433269 position adding 10176.824211 collateral
InvariantLib checking margin reqs for 0x7542002642420d2eea9164caa79a536dee18ae7f current 2.059819 + pending 0
  PositionLib calculating collateral reqd for position 2.059819 * price 50000.000000 * ratio .100000
  PositionLib checking whether collateral 19755.139196 >=? margin 10299.095000
```

Indeed, if I don't expect the small redeem to fail and let it run, the liquidation fails because maintenance requirements are no longer violated.

kbrizzle (approved):

PR #319, opened 2024-04-27
Add delegated signing

kbrizzle:
Adds `signer` role in parallel to `operator` that a user can assign to their account. Like operators, there can be many signers on one account, and signers can be revoked at any time.

A signer is able to sign messages on behalf of the user.

### DIFF
--- a/packages/perennial/contracts/MarketFactory.sol
+++ b/packages/perennial/contracts/MarketFactory.sol
@@ -24,6 +24,9 @@ contract MarketFactory is IMarketFactory, Factory {
     /// @dev The referreral fee level for each referrer
     mapping(address => UFixed6) public referralFee;
 
+    /// @dev Mapping of allowed signers for each account
+    mapping(address => mapping(address => bool)) public signers;
+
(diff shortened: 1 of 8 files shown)


packages/perennial/contracts/MarketFactory.sol
```diff
@@ -60,6 +63,14 @@ contract MarketFactory is IMarketFactory, Factory {
         emit OperatorUpdated(msg.sender, operator, newEnabled);
     }
 
+    /// @notice Updates the status of a signer for the caller
+    /// @param signer The signer to update
+    /// @param newEnabled The new status of the opersignerator
+    function updateSigner(address signer, bool newEnabled) external {
```
arjun-io, line 69: Will we need a `updateSignerWithSignature`?

arjun-io (approved):

packages/perennial/contracts/MarketFactory.sol
kbrizzle: There's a follow up task for that

PR #320, opened 2024-04-29
[v2.2] Delegate multiinvoker

arjun-io:
Enables a user to `updateOperator` an address to send multiinvoker transactions on their behalf via the `invoke(account, actions)` function

This is not fully gasless, as it still requires the user to send a transaction to approve the operator to act on their behalf

USDC or DSU is always pulled from the `account` (not `msg.sender`). The `msg.sender` is expected to provide the ETH update fee required for `COMMIT` actions


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -110,9 +113,33 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
+    /// @param account Account to perform invocations for
+    /// @param invocations List of actions to execute in order
+    function invoke(address account, Invocation[] calldata invocations) external payable {
+        _invoke(account, invocations);
+    }
+
+    /// @notice Performs a batch of invocations for an account
+    /// @param account Account to perform invocations for
+    /// @param invocations List of actions to execute in order
+    function _invoke(address account, Invocation[] calldata invocations) private {
```
kbrizzle, line 140: Little easier to follow if you name this `sender` instead of `account`, then it doesn't collide with `account` params as well.

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -47,6 +47,9 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @dev State for the order data
     mapping(address => mapping(IMarket => mapping(uint256 => TriggerOrderStorage))) private _orders;
 
+    /// @dev Approvals mapping to allow for sender to operate on behalf of account
+    mapping(address => mapping(address => bool)) public approvals;
```
kbrizzle, line 51: What do you think about naming this the same as the operator flow?
- `operators`
- `updateOperator()`
- `OperatorUpdated`

packages/perennial-extensions/contracts/MultiInvoker.sol
arjun-io: I feel like `sender` might also be confusing since the sender could be different from the account being operated on. This naming is the same as the operator + account flow we have in markets (i.e. an operator specifies the `account` to operate on in `update`)

kbrizzle (approved):

EdNoepel (approved):
LGTM

PR #323, opened 2024-05-03
Add solver fee

kbrizzle:
Adds a `solver` fee to intents.

Updates the `Intent` message to contain three new fields:
- `originator` the originator of the order (ex. the interface)
  - receives the subtractive fee less the solver fee
- `solver` the solver of the guarentee (ex. router API)
   - receives the solver fee
- `fee` portion of the subtractive fee to instead give to the solver.

All referral information is now read from the intent instead of passed in as a parameter to `update()`.

### DIFF
--- a/packages/perennial/contracts/Market.sol
+++ b/packages/perennial/contracts/Market.sol
@@ -82,7 +82,10 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     mapping(address => mapping(uint256 => address)) public liquidators;
 
     /// @dev The referrer for each id for each account
-    mapping(address => mapping(uint256 => address)) public referrers;
+    mapping(address => mapping(uint256 => address)) public orderReferrers;
+
+    /// @dev The referrer for each id for each account
+    mapping(address => mapping(uint256 => address)) public guaranteeReferrers;
 
     /// @dev The global pending guarantee for each id
     mapping(uint256 => GuaranteeStorageGlobal) private _guarantee;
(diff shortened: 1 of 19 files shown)


EdNoepel (approved):
I don't see any code issue.  Functionally, however, I'm curious why the actor submitting the intent gets to choose how much of the subtractive fee goes to the solver.

packages/perennial/contracts/Market.sol
```diff
@@ -475,8 +507,10 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         updateContext.operator = context.account == msg.sender || IMarketFactory(address(factory())).operators(context.account, msg.sender);
         updateContext.signer = context.account == signer || IMarketFactory(address(factory())).signers(context.account, signer);
         updateContext.liquidator = liquidators[context.account][context.local.currentId];
-        updateContext.referrer = referrers[context.account][context.local.currentId];
-        updateContext.referralFee = IMarketFactory(address(factory())).referralFee(referrer);
+        updateContext.orderReferrer = orderReferrers[context.account][context.local.currentId];
```
arjun-io, line 510: Do we need to clear these values when currentID increments like we did for the liquidator?

kbrizzle: Ahh yeah, these weren't included because we haven't merged this v2.2 fix in yet to v2.3.

I updated this, but there's going to be new merge conflict on your v2.2 -> v2.3 branch now.

kbrizzle:
@EdNoepel My idea was to have the solver fee be a cut of the interface fee.

This let's us piggy back on the "subtractive" nature of the interface fee without having to implement a whole parallel flow to keep track of two different referral fees.

I imagine this plays out where the solver API will only relay signed intents that include the proper fee they specify (like old school 0x orders). Then interfaces can choose which API they use and properly bundle the signed intents to comply with their fee schedule.

EdNoepel:
> @EdNoepel My idea was to have the solver fee be a cut of the interface fee.
> 
> This let's us piggy back on the "subtractive" nature of the interface fee without having to implement a whole parallel flow to keep track of two different referral fees.
Agree; much better than two fees.

> I imagine this plays out where the solver API will only relay signed intents that include the proper fee they specify (like old school 0x orders). Then interfaces can choose which API they use and properly bundle the signed intents to comply with their fee schedule.
Perhaps the SDK could auto-populate this upon connection to the relayer.

PR #344, opened 2024-05-24
Collateral accounts - market transfers

EdNoepel:
# Design

Two designs were considered.  In both designs, _Account_ contains token transfer logic, and _Controller_ contains logic required for rebalancing.

- In **design A**, _Account_ is market-agnostic, having two sets of functions for transferring tokens: one between the owner and account, and another between the account and controller.  The controller handles all market interactions, pulling funds into the controller and then into the market.
- In **design B**, _Account_ offers facilities for interacting with a single market, and the controller employs these facilities to perform transfers to market interactions.  User funds never touch the controller other than for purposes of keeper compensation.

Design **A** helped achieve a structural goal of keeping the _Account_ component "simple", but had several disadvantages:
- Additional token transfers were needed to move funds, increasing gas costs.
- _Controller_ needed to be approved as an operator on everyone's account.
- _Controller_ requires additional token approvals to interact with markets.

In addition to eliminating these disadvantages, design **B** offers some benefits.  To rebalance, _Controller_ only needs to read current state and make a single _Account_ call for each market being rebalanced.  This simplifies code maintainance while maintaining a well-defined separation of concerns between components and balancing contract size.

## Recent changes
In _Account_, `wrap` and `unwrap` facilities were made public.  Owners do not need to interact with these methods.  They were exposed such that the controller can implicitly wrap funds to compensate keepers during account creation and when assigning delegates before the account is used.

# TODO
- [x] Review coverage and add tests where appropriate
- [x] Resolve anticipated CI failures, such as tests failing under coverage
- [x] Examine gas required to add custom revert if insufficient funds for keeper compensation, rather than an ambiguous ERC20 transfer error
- [ ] Self-review and resolve linter warnings


packages/perennial-account/contracts/Account.sol
```diff
@@ -60,33 +65,70 @@ contract Account is IAccount {
...
+        // handle magic number for full withdrawal
+        if (amount.eq(Fixed6Lib.MIN)){
+            // ensure user has a positive collateral balance to withdraw
+            Fixed6 balance = market.locals(owner).collateral;
+            if (balance.sign() != 1) revert NoCollateral(address(market));
+            // ensure user has no position
+            amount = balance.mul(Fixed6Lib.NEG_ONE);
+
+        // handle magic number for full deposit
+        } else if (amount.eq(Fixed6Lib.MAX)){
```
kbrizzle, line 83: I don't think this magic value / clause is necessary from a UX point of view.

"Withdraw all" is a common use case, but usually on the deposit side the user will always want to specify and exact value because this determines their leverage in the market relative to their position.

packages/perennial-account/contracts/Controller_Arbitrum.sol
```diff
@@ -95,6 +118,16 @@ contract Controller_Arbitrum is Controller, Kept_Arbitrum {
         (address account, uint256 maxFee) = abi.decode(data, (address, uint256));
         // maxFee is a UFixed6; convert to 18-decimal precision
         raisedKeeperFee = amount.min(UFixed18.wrap(maxFee * 1e12));
-        keeperToken().pull(account, raisedKeeperFee);
+
+        // if the account has insufficient DSU to pay the fee, wrap
+        if (DSU.balanceOf(account).lt(raisedKeeperFee)) {
```
kbrizzle, line 123: Instead re-implementing the whole `if !balance ... wrap then pull` can we reuse the `withdraw` logic that's already handling this in `Account`, and maybe update it so it can also withdraw to the controller?

There would still be a little complexity there on access controller and how to allow a different destination cleanly / securely.

packages/perennial-account/contracts/Account.sol
```diff
@@ -60,33 +65,70 @@ contract Account is IAccount {
...
+        // implicitly approve this market to spend our DSU
+        uint256 allowance = IERC20Metadata(Token18.unwrap(DSU)).allowance(address(this), address(market));
+        if (allowance != type(uint256).max)
+            DSU.approve(address(market));
+
+        // handle magic number for full withdrawal
+        if (amount.eq(Fixed6Lib.MIN)){
+            // ensure user has a positive collateral balance to withdraw
+            Fixed6 balance = market.locals(owner).collateral;
+            if (balance.sign() != 1) revert NoCollateral(address(market));
```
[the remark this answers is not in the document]
EdNoepel: Renamed in b632eab7d4c3ca68155dc9d2299a98a7283689e1.

packages/perennial-account/contracts/Account.sol
EdNoepel: We have some other options to handle keeper compensation for market deposits.  Assume a user puts 2000 USDC in their account.
1.  Always deduct the fee from the deposit amount.  For example, if the user signs a message to to transfer 1000 DSU into a market, that the market would have 999.7 DSU.
2. Check balances and only meddle with the deposit amount if there are insufficient funds to pay the keeper.
3. Simply revert if the user specifies a full deposit.  This UX seems undesirable.

Option 2 seems awkward yet the best UX.  Your thoughts would be appreciated.

EdNoepel: I attempted to implement option 2, but meddling with the deposit amount within the controller was predictably awkward.  A fourth proposal:

4. Limit market deposits by the (post-wrap, post-keeper-fee) account balance.

This seems a more elegant way to compensate the keeper when user explicitly requests a deposit amount equal to their account balance.  I implemented this in [ed/collateral-accounts_transfer-no-magic-full-deposit](https://github.com/equilibria-xyz/perennial-v2/commit/c2c957798ad7959ec0c9c0a6b5f5f503d86c780b).  @kbrizzle Please have a look; I'll merge into this PR if you like this solution.

packages/perennial-account/contracts/Account.sol
```diff
@@ -60,33 +65,70 @@ contract Account is IAccount {
     }
 
     /// @inheritdoc IAccount
-    function withdraw(UFixed6 amount, bool unwrap) external ownerOrController {
+    function marketTransfer(IMarket market, Fixed6 amount) external ownerOrController {
+        // implicitly approve this market to spend our DSU
+        uint256 allowance = IERC20Metadata(Token18.unwrap(DSU)).allowance(address(this), address(market));
+        if (allowance != type(uint256).max)
```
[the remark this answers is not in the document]
EdNoepel: I added a test which deposits to the same market multiple times.  The gas benefit is quite small:
```
with allowance check, 12 deposits to same market
·-------------------------------------------------|---------------------------|-----------------|-----------------------------·
|              Solc version: 0.8.24               ·  Optimizer enabled: true  ·  Runs: 1000000  ·  Block limit: 30000000 gas  │
··················································|···························|·················|······························
|  Methods                                                                                                                    │
··················|·······························|·············|·············|·················|···············|··············
|  Contract       ·  Method                       ·  Min        ·  Max        ·  Avg            ·  # calls      ·  usd (avg)  │
··················|·······························|·············|·············|·················|···············|··············
|  Controller     ·  marketTransferWithSignature  ·     624965  ·    1082979  ·         663142  ·           36  ·          -  │
··················|·······························|·············|·············|·················|···············|··············

without allowance check, 12 deposits to same market
·-------------------------------------------------|---------------------------|-----------------|-----------------------------·
|              Solc version: 0.8.24               ·  Optimizer enabled: true  ·  Runs: 1000000  ·  Block limit: 30000000 gas  │
··················································|···························|·················|······························
|  Methods                                                                                                                    │
··················|·······························|·············|·············|·················|···············|··············
|  Contract       ·  Method                       ·  Min        ·  Max        ·  Avg            ·  # calls      ·  usd (avg)  │
··················|·······························|·············|·············|·················|···············|··············
|  Controller     ·  marketTransferWithSignature  ·     626416  ·    1081332  ·         664336  ·           36  ·          -  │
··················|·······························|·············|·············|·················|···············|··············

```
I removed the allowance check in 9e791827af9fe1e327ad9884f75b310ce1183f3f.

packages/perennial-account/contracts/Account.sol
```diff
@@ -60,33 +65,70 @@ contract Account is IAccount {
...
+            DSU.approve(address(market));
+
+        // handle magic number for full withdrawal
+        if (amount.eq(Fixed6Lib.MIN)){
```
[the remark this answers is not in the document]
EdNoepel: Done in 4e02261abced314dbda17f244e535bafe79c33fa.

packages/perennial-account/contracts/Account.sol
kbrizzle: Looking through these, I actually think (3) is the safest from a preventing footguns perspective.

Consider a user who wants to deposit $1000 and then open 10x ETH long
 - If our cap was hit via (4), and say only $100 was deposited (extreme example), our trader would be at 100x leverage instead of 10x once the position opened.

I think it's better to have these either fully fill at once or don't execute at all. As a follow up we need to make sure intents have some mechanism to "only open w/ sufficient collateral", but i'll make a separate task for that.

EdNoepel: Done in 249064622de9cf0fa97dd1acb32f2e4a73b3cd8d.

packages/perennial-account/contracts/Controller_Arbitrum.sol
EdNoepel: `withdraw` handles transferring USDC out of the account; it has an _unwrap_ flow, while `_raiseKeeperFee` needs a _wrap_ flow.

kbrizzle (approved):

PR #354, opened 2024-06-12
[Fix] riskParameter migration compatibility

kbrizzle:
Ensures that the latest -> new storage accessors properly work in `updateRiskParameter` during the v2.3 migration process.

- Cleans up the `updateRiskParameter` logic, to require minimal backwards compatibility.
- Updates `RiskParameter` layout to ensure `takerFee.adiabatic / takerFee.scale` are backwards compatible.

### DIFF
--- a/packages/perennial/contracts/types/Global.sol
+++ b/packages/perennial/contracts/types/Global.sol
@@ -42,6 +42,22 @@ using GlobalStorageLib for GlobalStorage global;
 /// @title Global
 /// @notice Holds the global market state
 library GlobalLib {
+    /// @notice Updates market exposure based on a change in the risk parameter configuration
+    /// @param self The Global object to update
+    /// @param latestRiskParameter The latest risk parameter configuration
+    /// @param newRiskParameter The new risk parameter configuration
+    /// @param latestPosition The latest position
+    function update(
+        Global memory self,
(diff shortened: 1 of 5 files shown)


packages/perennial/contracts/types/Global.sol
```diff
@@ -42,6 +42,22 @@ using GlobalStorageLib for GlobalStorage global;
...
+    /// @param latestPosition The latest position
+    function update(
+        Global memory self,
+        RiskParameter memory latestRiskParameter,
+        RiskParameter memory newRiskParameter,
+        Position memory latestPosition
+    ) internal pure {
+        Fixed6 exposureChange = latestRiskParameter.takerFee
+            .exposure(newRiskParameter.takerFee, latestPosition.skew(), self.latestPrice.abs());
+        self.exposure = self.exposure.sub(exposureChange);
```
arjun-io, line 58: Is this external safe or do we need to return the global?

arjun-io (approved):
One question

packages/perennial/contracts/types/Global.sol
kbrizzle: Only the `/libs` dir and the storage type libs are external safe, the logic type libs are not external safe.

Will be clarifying this in a future task.

PR #355, opened 2024-06-13
Collateral Accounts - rebalance configuration

EdNoepel:
Implemented messaging and storage for rebalance group configuration.  Actual rebalancing will be implemented in another PR.

## TODO
- [x] Confirm CI runs cleanly.
- [x] Check test coverage.
- [x] Resolve linter warnings.


packages/perennial-account/contracts/Controller.sol
```diff
@@ -38,11 +49,19 @@ contract Controller is Instance, IController {
...
     mapping(address => mapping(address => bool)) public signers;
 
-    /// @notice Configures the EIP-712 message verifier used by this controller
-    /// @param verifier_ Contract used to validate messages were signed by the sender
-    /// @param usdc_ USDC token address
-    /// @param dsu_ DSU token address
-    /// @param reserve_ DSU Reserve address, used by Account
+    /// @dev Mapping of rebalance configuration
+    /// owner => group => market => config
+    mapping(address => mapping(uint8 => mapping(address => RebalanceConfig))) config;
```
kbrizzle, line 54: I think it's actually extra gas / code size to use `uint8` here instead of `uint256` because it adds casting / overflow checks throughout.

packages/perennial-account/contracts/Controller.sol
```diff
@@ -118,6 +137,48 @@ contract Controller is Instance, IController {
...
+        _changeRebalanceConfigWithSignature(configChange, signature);
+    }
+
+    function _changeRebalanceConfigWithSignature(RebalanceConfigChange calldata configChange, bytes calldata signature) internal {
+        // ensure the message was signed by the owner or a delegated signer
+        verifier.verifyRebalanceConfigChange(configChange, signature);
+        _ensureValidSigner(configChange.action.common.account, configChange.action.common.signer);
+
+        // sum of the target allocations of all markets in the group
+        UFixed6 totalAllocation;
```
kbrizzle, line 154: style: no need to initialize this first on a separate line.

packages/perennial-account/contracts/Controller.sol
```diff
@@ -118,6 +137,48 @@ contract Controller is Instance, IController {
...
+        // put this on the stack for readability
+        address owner = configChange.action.common.account;
+
+        totalAllocation = _updateRebalanceGroup(configChange, owner);
+
+        // if not deleting the group, ensure rebalance targets add to 100%
+        if (configChange.markets.length != 0 && !totalAllocation.eq(RebalanceConfigLib.MAX_PERCENT))
```
kbrizzle, line 161: I think it's ok to just use `UFixed6Lib.ONE` instead of making a new constant for max percent (we do this throughout the code where 100% comes up frequently).

packages/perennial-account/contracts/Controller.sol
```diff
@@ -161,4 +222,54 @@ contract Controller is Instance, IController {
...
+    /// @dev overwrites rebalance configuration of all markets for a particular owner and group
+    /// @param message already-verified message with new configuration
+    /// @param owner identifies the owner of the collateral account
+    function _updateRebalanceGroup(
+        RebalanceConfigChange calldata message,
+        address owner
+    ) private returns (UFixed6 totalAllocation) {
+        // ensure group index is valid
+        if (message.group == 0 || message.group > MAX_GROUPS_PER_OWNER)
+            revert IController.ControllerInvalidRebalanceGroup();
```
kbrizzle, line 235: I don't think you need the `IController.` here in a few places

packages/perennial-account/contracts/Controller.sol
```diff
@@ -161,4 +222,54 @@ contract Controller is Instance, IController {
...
+
+    /// @dev overwrites rebalance configuration of all markets for a particular owner and group
+    /// @param message already-verified message with new configuration
+    /// @param owner identifies the owner of the collateral account
+    function _updateRebalanceGroup(
+        RebalanceConfigChange calldata message,
+        address owner
+    ) private returns (UFixed6 totalAllocation) {
+        // ensure group index is valid
+        if (message.group == 0 || message.group > MAX_GROUPS_PER_OWNER)
```
kbrizzle, line 234: why can we not use group zero with the new data structure?

packages/perennial-account/contracts/Controller.sol
```diff
@@ -161,4 +222,54 @@ contract Controller is Instance, IController {
...
+            revert IController.ControllerInvalidRebalanceGroup();
+
+        if (message.markets.length > MAX_MARKETS_PER_GROUP)
+            revert IController.ControllerInvalidRebalanceMarkets();
+
+        // delete the existing group
+        for (uint256 i; i < groupToMarkets[owner][message.group].length; ++i) {
```
kbrizzle, line 241: `i++` should be the same gas cost now

packages/perennial-account/contracts/Controller.sol
kbrizzle: Also, I think it's a little more readable to encapsulate this invariant within the `_updateRebalanceGroup` function

packages/perennial-account/contracts/Controller.sol
```diff
@@ -161,4 +222,54 @@ contract Controller is Instance, IController {
...
+                revert IController.ControllerMarketAlreadyInGroup(message.markets[i], currentGroup);
+
+            // rewrite over all the old configuration
+            marketToGroup[owner][message.markets[i]] = message.group;
+            config[owner][message.group][message.markets[i]] = message.configs[i];
+            groupToMarkets[owner][message.group].push(message.markets[i]);
+
+            // ensure target allocation across all markets totals 100%
+            // read from storage to trap duplicate markets in the message
+            totalAllocation = totalAllocation.add(
```
kbrizzle, line 261: style: we'd usually write this like this:
```
            totalAllocation = totalAllocation
                .add(config[owner][message.group][message.markets[i]].target);
```

packages/perennial-account/contracts/Controller.sol
```diff
@@ -161,4 +222,54 @@ contract Controller is Instance, IController {
...
+                config[owner][message.group][message.markets[i]].target
```
kbrizzle, line 262: I think you want to do `message.configs[i].target` here instead to avoid the extra sload (even though it's already warm).

packages/perennial-account/contracts/Controller.sol
```diff
@@ -161,4 +222,54 @@ contract Controller is Instance, IController {
...
+            address market = groupToMarkets[owner][message.group][i];
+            delete config[owner][message.group][market];
+            delete marketToGroup[owner][market];
+        }
+        delete groupToMarkets[owner][message.group];
+
+        for (uint256 i; i < message.markets.length; ++i) {
+            // ensure market is not pointing to a different group
+            uint8 currentGroup = marketToGroup[owner][message.markets[i]];
+            if (currentGroup != 0)
```
kbrizzle, line 251: I would just inline `marketToGroup[owner][message.markets[i]]` here since we're never using `currentGroup` again.

packages/perennial-account/contracts/types/RebalanceConfig.sol
```diff
@@ -0,0 +1,106 @@
...
+            uint256(UFixed6.unwrap(newValue.threshold) << (256 - 32)) >> (256 - 32 - 32);
+
+        assembly {
+            sstore(self.slot, encoded0)
+        }
+    }
+}
+
+/// @dev Action message to change configuration for a group of markets
+struct RebalanceConfigChange {
```
kbrizzle, line 61: I think that you should separate out this into another type file for readability.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Markets deleted from or never added to `marketToGroup` will always be reported as belonging to group 0.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Indeed; when implementing it I measured it cost an extra 0.725%.  Thought you would want it for readability.  Reverted the commit in a21dcde67987088ca5c4470e7bd08d8319dcf42f.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Agree; done in 730cf71379aa973b26dea9c964080cb3ea825ea4.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Did this for readability, since we include the group in the revert params in the next line.  I'll do it if you insist, but the revert would look ugly.

packages/perennial-account/contracts/Controller.sol
EdNoepel: That's embarassing; I refactored this method when doing the keeper compensation and forgot to clean it up afterward.  Fixed in 6f695f61d1888d27adc2b948582f8ce690c9fc10.

packages/perennial-account/contracts/Controller.sol
EdNoepel: No problem; done in 6f695f61d1888d27adc2b948582f8ce690c9fc10.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Changed in b29b327669cf01c0b25f4ac0d2d8a154cc5dd44a.

packages/perennial-account/contracts/Controller.sol
EdNoepel: The whole thing fits on one line now after reading config from message.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Agree; done in 730cf71379aa973b26dea9c964080cb3ea825ea4.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Removed in 37eeb0fff4135ff48f220d822e54d737915bda5c.

packages/perennial-account/contracts/types/RebalanceConfig.sol
EdNoepel: Moved to its own file in e88366467b394f265fa3ad0286ce1896d6ef203d.

packages/perennial-account/contracts/Controller.sol
kbrizzle: Fair point on readability. Since this issue is so widespread, we default to `uint256` for all in-memory primitive numbers.

EdNoepel:
> we've kind of landed on a style of not emitting any parameters with custom reverts. Just from a practicality perspective, we've found that they rarely end up being useful over other debugging options. Do you have a strong opinion here on including these in the revert?

I agree error parameters are not helpful when there is a clear relationship between inputs and the cause of the issue.  However, when the relationship between user inputs and the issue is detached, parameters can provide value.  The usual use case involves working with arrays.  Identifying the item causing a problem improves DX/UX.  Here's my rationale for the two places I'm providing parameters:

**ControllerMarketAlreadyInGroupError** - When configuring a group with 4 markets, the _market_ parameter shows which entry is the problem, and the _group_ parameter shows the user where to remove it (if desired).

**ControllerUnsupportedMarketError** - When groups are configured, we do not validate the `IMarket`.  A problem may not be discovered until the user attempts a rebalance.  Assuming there are other markets in the group, the _market_ parameter identifies which market is incompatible with collateral accounts, such that the user may remove it.

packages/perennial-account/contracts/Controller.sol
```diff
@@ -161,4 +210,57 @@ contract Controller is Instance, IController {
...
+                owner,
+                message.group,
+                message.markets[i],
+                message.configs[i]
+            );
+        }
+
+        // if not deleting the group, ensure rebalance targets add to 100%
+        if (message.markets.length != 0 && !totalAllocation.eq(UFixed6Lib.ONE))
+            revert ControllerInvalidRebalanceTargets();
```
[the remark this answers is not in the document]
EdNoepel: done in b5d3948e25be6b350fcf09ad3b8a966c74fee190

kbrizzle (approved):

PR #363, opened 2024-06-19
Add newPrice option oracle request

kbrizzle:
Adds a `newPrice` boolean argument to the `oracleProvider.request()` function which allows the end consumer to specify whether they would like a fresh price at the current timestamp, or whether it is acceptable to carry forward the latest previous price.

In either case a price is guaranteed to be populated at the current version in order as expected, however requests that do not require a fresh price do not need to pay for the settlement fee they do not trigger a keeper callback directly.

This allows us to remove all price-override logic from the market because we can ensure that a price is available for any case of market update.


EdNoepel (approved):
LGTM.  I'm currently reworking the collateral accounts oracle setup in another branch.

PR #365, opened 2024-06-24
Colateral Accounts - rebalance logic

EdNoepel:
# Changes
Added implementation logic to:
- Check whether a group may be rebalanced (intended to be called offchain).
- Perform a rebalance for a particular group.

Also:
- Tidy comments and sort _Controller_ methods.
- Corrected error signatures for recently-renamed errors.

# TODO
- [x] Confirm CI runs cleanly
- [x] Check coverage
- [x] Self re-review

### DIFF
--- a/packages/perennial-account/contracts/Controller.sol
+++ b/packages/perennial-account/contracts/Controller.sol
@@ -236,14 +236,14 @@ contract Controller is Instance, IController {
     }
 
     function _rebalanceGroup(address owner, uint256 group) internal {
-        // query owner's collateral in each market and calculate sum
-        (Fixed6[] memory actualCollateral, Fixed6 groupCollateral) = _queryMarketCollateral(owner, group);
-        IAccount account = IAccount(getAccountAddress(owner));
+        // settles each markets, such that locals are up-to-date
+        _settleMarkets(owner, group);
 
         // determine imbalances
         (, bool canRebalance, Fixed6[] memory imbalances) = checkGroup(owner, group);
         if (!canRebalance) revert ControllerGroupBalancedError();
 
+        IAccount account = IAccount(getAccountAddress(owner));
         // pull collateral from markets with surplus collateral
         for (uint256 i; i < imbalances.length; i++) {
             IMarket market = groupToMarkets[owner][group][i];
@@ -278,15 +278,23 @@ contract Controller is Instance, IController {
         if (signer != owner && !signers[owner][signer]) revert ControllerInvalidSignerError();
     }
 
+    /// @dev checks current collateral for each market in a group and aggregates collateral for the group
     function _queryMarketCollateral(address owner, uint256 group) private view returns (
         Fixed6[] memory actualCollateral,
         Fixed6 groupCollateral
     ) {
         actualCollateral = new Fixed6[](groupToMarkets[owner][group].length);
-        for (uint256 i; i < groupToMarkets[owner][group].length; i++) {
+        for (uint256 i; i < actualCollateral.length; i++) {
             Fixed6 collateral = groupToMarkets[owner][group][i].locals(owner).collateral;
             actualCollateral[i] = collateral;
             groupCollateral = groupCollateral.add(collateral);
+        }
+    }
+
+    /// @dev settles each market in a rebalancing group
+    function _settleMarkets(address owner, uint256 group) private {
+        for (uint256 i; i < groupToMarkets[owner][group].length; i++) {
+            groupToMarkets[owner][group][i].settle(owner);
         }
     }
 
--- a/packages/perennial-account/contracts/Controller.sol
+++ b/packages/perennial-account/contracts/Controller.sol
@@ -250,7 +250,7 @@ contract Controller is Instance, IController {
             }
         }
 
-        if (!canAnyMarketRebalance) revert ControllerGroupBalanced();
+        if (!canAnyMarketRebalance) revert ControllerGroupBalancedError();
 
         // push collateral to markets with insufficient collateral
         for (uint256 i; i < imbalances.length; i++) {
--- a/packages/perennial-account/contracts/Controller.sol
+++ b/packages/perennial-account/contracts/Controller.sol
@@ -25,6 +25,7 @@ import {
 } from "./types/RebalanceConfig.sol";
 import { SignerUpdate, SignerUpdateLib } from "./types/SignerUpdate.sol";
 import { Withdrawal, WithdrawalLib } from "./types/Withdrawal.sol";
+import "hardhat/console.sol";
 
 /// @title Controller
 /// @notice Facilitates unpermissioned actions between collateral accounts and markets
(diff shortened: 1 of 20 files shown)


packages/perennial-account/contracts/interfaces/IController.sol
```diff
@@ -46,38 +52,42 @@ interface IController {
     /// @param newEnabled True to assign or enable, false to disable
     event SignerUpdated(address indexed owner, address indexed signer, bool newEnabled);
 
-    // sig: 0x2c51df8b
+    // sig: 0x599f7892
     /// @custom:error Insufficient funds in the collateral account to compensate relayer/keeper
     error ControllerCannotPayKeeperError();
 
-    // sig: 0x1444bc5d
+    /// sig: 0xdc72f280
```
EdNoepel, line 59: Will fix to `//` in the next push.

packages/perennial-account/contracts/Controller.sol
```diff
@@ -206,11 +231,66 @@ contract Controller is Instance, IController {
...
 
+    function _rebalanceGroup(address owner, uint256 group) internal {
+        // query owner's collateral in each market and calculate sum
+        (Fixed6[] memory actualCollateral, Fixed6 groupCollateral) = _queryMarketCollateral(owner, group);
+        IAccount account = IAccount(getAccountAddress(owner));
+
+        // create an array with imbalances, pull collateral from markets with surplus collateral
+        Fixed6[] memory imbalances = new Fixed6[](actualCollateral.length);
+        bool canAnyMarketRebalance;
+        for (uint256 i; i < actualCollateral.length; i++) {
```
kbrizzle, line 242: can we combine this logic with `checkGroup` somehow? Maybe something like

- `checkGroup` computes 1) actual collateral 2) target collateral 3) boolean for rebalancable
- `_rebalanceGroup` if boolean is true, runs through list and does the market transfers per difference in 1 + 2

I think this simplifies the code and maybe saves some complexity from the `RebalanceLib`, but what do you think?

packages/perennial-account/contracts/Controller.sol
```diff
@@ -206,11 +231,66 @@ contract Controller is Instance, IController {
...
+        signers[owner][signerUpdate.signer] = signerUpdate.approved;
+        emit SignerUpdated(owner, signerUpdate.signer, signerUpdate.approved);
+    }
+
     /// @dev calculates the account address and reverts if user is not authorized to sign transactions for the owner
     function _ensureValidSigner(address owner, address signer) private view {
         if (signer != owner && !signers[owner][signer]) revert ControllerInvalidSignerError();
     }
 
+    function _queryMarketCollateral(address owner, uint256 group) private view returns (
```
kbrizzle, line 281: you'll need to settle the markets prior to this other the locals won't be up to date (you can assume the market update invariant will handle the settlement being "fresh")

packages/perennial-account/contracts/types/RebalanceConfigChange.sol
```diff
@@ -15,6 +15,8 @@ struct RebalanceConfigChange {
     address[] markets;
     /// @dev Target allocation for markets in the aforementioned array
     RebalanceConfig[] configs;
+    /// @dev Largest amount to compensate a relayer/keeper for rebalancing the group in DSU
+    UFixed6 maxRebalanceFee;
```
kbrizzle, line 19: naming: you can just call this `maxFee`

packages/perennial-account/contracts/Controller.sol
```diff
@@ -206,11 +231,66 @@ contract Controller is Instance, IController {
...
+        Fixed6[] memory actualCollateral,
+        Fixed6 groupCollateral
+    ) {
+        actualCollateral = new Fixed6[](groupToMarkets[owner][group].length);
+        for (uint256 i; i < groupToMarkets[owner][group].length; i++) {
+            IMarket market = IMarket(groupToMarkets[owner][group][i]);
```
kbrizzle, line 287: You can make the type in the mapping `IMarket` so you don't need to do the casting here, then it's easier to one-line as well.

packages/perennial-account/contracts/Controller.sol
```diff
@@ -181,23 +165,64 @@ contract Controller is Instance, IController {
...
+        MarketTransfer calldata marketTransfer,
+        bytes calldata signature
+    ) internal {
+        // ensure the message was signed by the owner or a delegated signer
+        verifier.verifyMarketTransfer(marketTransfer, signature);
+        _ensureValidSigner(marketTransfer.action.common.account, marketTransfer.action.common.signer);
+
+        // only Markets with DSU collateral are supported
+        IMarket market = IMarket(marketTransfer.market);
+        if (!market.token().eq(DSU)) revert ControllerUnsupportedMarketError(address(market));
```
kbrizzle, line 216: you can use `IMarket` type in the error / events to avoid the casting here.

packages/perennial-account/contracts/types/RebalanceConfigChange.sol
EdNoepel: Done in ab298615400426aa18ac74e4c5631b67679243f2.

packages/perennial-account/contracts/Controller.sol
EdNoepel: Done in 338410d364315a300efe592569ec434da51e5792.

packages/perennial-account/contracts/Controller.sol
EdNoepel: For consistency, that means also updating `ControllerMarketAlreadyInGroupError` error to provide the interface rather than the address, and casting there instead.  Done in 338410d364315a300efe592569ec434da51e5792.

packages/perennial-account/contracts/Controller.sol
EdNoepel: We need to consider both the state the keeper sees and the state upon update.  A few use cases:
A) **One or more markets have settled and collateral is outside defined thresholds.  Locals will remain outside thresholds in the next version.** - The keeper will see an opportunity and should be able to successfully rebalance.
B) **Collateral in all markets are currently within thresholds, but one or more markets will be outside defined thresholds in the next version.** - The check view does not predict what locals will be after the next settlement, so keepers cannot see such opportunities.
C) **Collateral in one or more markets is currently outside of thresholds, but settlement will bring locals back within thresholds.** - The keeper will see an opportunity.  Leaving code as-is, the rebalance would succeed, but be unnecessary.  If we settle first, the keeper could be compensated for settlement and avoid rebalancing.

The current implementation only handles case A).

I could refactor `_queryMarketCollateral` such that it is no longer shared between the check and the action, settling each market in the latter case, prior to the rebalance.  This would give us more accurate locals for handling cases A) and C), but it will not help with case B).  Is this the way, or do we need to somehow handle case B?

packages/perennial-account/contracts/Controller.sol
EdNoepel: Yes; in 1197527bd972abeb07b91b8e8fc12aed56bc6bbe I updated `checkGroup` to store and return the imbalances.  The biggest win is improving the readability/auditability of `_rebalanceGroup`.

Small gas increase, but seems worth it:
```
·--------------------------------------------------------------|---------------------------|-----------------|-----------------------------·
| Before              Solc version: 0.8.24                     ·  Optimizer enabled: true  ·  Runs: 1000000  ·  Block limit: 32000000 gas  │
·······························································|···························|·················|······························
|  Methods                                                                                                                                 │
························|······································|·············|·············|·················|···············|··············
|  Contract             ·  Method                              ·  Min        ·  Max        ·  Avg            ·  # calls      ·  usd (avg)  │
························|······································|·············|·············|·················|···············|··············
|  Controller_Arbitrum  ·  rebalanceGroup                      ·    1533287  ·    1533302  ·        1533295  ·           10  ·          -  │
························|······································|·············|·············|·················|···············|··············
|  Controller           ·  rebalanceGroup                      ·     888494  ·    1456435  ·        1172465  ·            8  ·          -  │
························|······································|·············|·············|·················|···············|··············


·--------------------------------------------------------------|---------------------------|-----------------|-----------------------------·
| After               Solc version: 0.8.24                     ·  Optimizer enabled: true  ·  Runs: 1000000  ·  Block limit: 32000000 gas  │
·······························································|···························|·················|······························
|  Methods                                                                                                                                 │
························|······································|·············|·············|·················|···············|··············
|  Controller_Arbitrum  ·  rebalanceGroup                      ·    1547845  ·    1547860  ·        1547853  ·           10  ·          -  │
························|······································|·············|·············|·················|···············|··············
|  Controller           ·  rebalanceGroup                      ·     902985  ·    1470926  ·        1186956  ·            8  ·          -  │
························|······································|·············|·············|·················|···············|··············
```

I have been on the fence about whether `RebalanceLib` is needed.  It seemed wise to keep rebalancing logic in an external lib to minimize contract size.  But since we cannot compartmentalize configuration storage into a struct, there isn't much logic we can offload to the library.  Maybe we don't need it.  My plan is to wait until UI/keeper requirements are understood before removing.

packages/perennial-account/contracts/Controller.sol
EdNoepel: My latest commit cleans this up, settling prior to rebalance.  Regarding use case B, we still rely on external settlement before a keeper sees an opportunity to rebalance.

kbrizzle: Keeper's in general virtually settle anyways before making getter calls (because this settlement issue exists all over), so this shouldn't be an issue in practice.

kbrizzle (approved):

PR #367, opened 2024-07-02
Collateral accounts: deploy accounts to a BeaconProxy

EdNoepel:
https://github.com/equilibria-xyz/root/pull/97 adds _create2_ functionality to the existing `Factory` attribute, and is a dependency for this change.  Will need to update `package.json` once a new RC is published, such that everything (including CI) will compile and run.  Until then, can use `yarn link` to use a local checkout of that `root` branch.

This PR deterministically deploys _Accounts_ to a `BeaconProxy`, similar to how factories are used elsewhere in the protocol.  The proxy for the _Controller_ itself is to be handled as a deployment task, outside the scope of this PR.

### DIFF
--- a/packages/perennial-account/contracts/Account.sol
+++ b/packages/perennial-account/contracts/Account.sol
@@ -46,14 +46,11 @@ contract Account is IAccount, Instance {
         owner = owner_;
 
         // approve the Controller to interact with this account's DSU
-        if (address(factory()) != address(0))
-            DSU.approve(address(factory()));
+        DSU.approve(address(factory()));
 
         // approve DSU facilities to wrap and unwrap USDC for this account
-        if (address(reserve) != address(0)) {
-            DSU.approve(address(reserve));
-            USDC.approve(address(reserve));
-        }
+        DSU.approve(address(reserve));
+        USDC.approve(address(reserve));
     }
 
     /// @inheritdoc IAccount
--- a/packages/perennial-account/contracts/Account.sol
+++ b/packages/perennial-account/contracts/Account.sol
@@ -22,35 +22,37 @@ contract Account is IAccount, Instance {
     address public owner;
 
     /// @dev USDC stablecoin address
-    Token6 public USDC; // solhint-disable-line var-name-mixedcase
+    Token6 public immutable USDC; // solhint-disable-line var-name-mixedcase
 
     /// @dev DSU address
-    Token18 public DSU; // solhint-disable-line var-name-mixedcase
+    Token18 public immutable DSU; // solhint-disable-line var-name-mixedcase
 
     /// @dev DSU Reserve address
-    IEmptySetReserve public reserve;
+    IEmptySetReserve public immutable reserve;
 
-    /// @inheritdoc IAccount
-    function initialize(
-        address owner_,
-        Token6 usdc_,
-        Token18 dsu_,
-        IEmptySetReserve reserve_
-    ) external initializer(1) {
-        __Instance__initialize();
-        owner = owner_;
+    /// @dev Construct collateral account and set approvals for controller and DSU reserve
+    /// @param usdc_ USDC stablecoin
+    /// @param dsu_ Digital Standard Unit stablecoin
+    /// @param reserve_ DSU SimpleReserve contract, used for wrapping/unwrapping USDC to/from DSU
+    constructor(Token6 usdc_, Token18 dsu_, IEmptySetReserve reserve_) {
         USDC = usdc_;
         DSU = dsu_;
         reserve = reserve_;
+    }
+
+    /// @inheritdoc IAccount
+    function initialize(address owner_) external initializer(1) {
+        __Instance__initialize();
+        owner = owner_;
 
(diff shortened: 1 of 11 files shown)


packages/perennial-account/contracts/Controller.sol
```diff
@@ -79,16 +83,8 @@ contract Controller is Instance, IController {
...
-            type(Account).creationCode,
-            abi.encode(owner),
-            abi.encode(address(this)),
-            abi.encode(USDC),
-            abi.encode(DSU),
-            abi.encode(reserve));
-        // calculate the hash for that bytecode and compute the address
-        return Create2.computeAddress(SALT, keccak256(bytecode));
+        // calculate the hash for an uninitialized account for the owner
+        return _computeCreate2Address("", keccak256(abi.encode(SALT, owner)));
```
EdNoepel, line 87: We don't pass initialization params here, because we only want the address salt to be affected by our constant and the owner.  For example, we shouldn't generate a new address if DSU address changes.

packages/perennial-account/contracts/Account.sol
```diff
@@ -3,60 +3,55 @@ pragma solidity 0.8.24;
...
-        address controller_,
         Token6 usdc_,
         Token18 dsu_,
         IEmptySetReserve reserve_
-    ) {
+    ) external initializer(1) {
+        __Instance__initialize();
         owner = owner_;
-        controller = controller_;
         USDC = usdc_;
```
kbrizzle, line 42: you can still have USDC, DSU, and reserve be immutable even though the others are set via the constructor.

- constructor is set when the implementation is deployed (owner)
- initializer is called when instance is deployed (USDC, DSU, reserve)

packages/perennial-account/contracts/Account.sol
```diff
@@ -3,60 +3,55 @@ pragma solidity 0.8.24;
...
         DSU = dsu_;
         reserve = reserve_;
 
-        dsu_.approve(controller_);
+        // approve the Controller to interact with this account's DSU
+        if (address(factory()) != address(0))
```
kbrizzle, line 47: I think you can leave off the address zero check in these two places - there's not really a huge harm here even if it is somehow incorrectly configured.

packages/perennial-account/contracts/Controller.sol
kbrizzle: Oh interesting, is that why you moved them all to storage instead of immutable?

packages/perennial-account/contracts/Account.sol
EdNoepel: When deploying the implementation, I supposed we would initialize using the 0x0 address to prevent another actor from calling it.  Approving the 0x0 address reverts, which is why I added the conditionals.  Is there another way we should protect the implementation?

EdNoepel: I see now that `Instance.sol` uses `onlyInitializer` on `__Instance__initialize()`, so this shouldn't be a problem.  Will adjust accordingly.

packages/perennial-account/contracts/Account.sol
EdNoepel: Should be other way around, right?  USDC, DSU, reserve could be specified upon deployment of the implementation and handled by constructor.  Owner needs to be set in `initialize` function.

packages/perennial-account/contracts/Account.sol
EdNoepel: No need for the zero checks now that stablecoins are set in the constructor; removed in d3ec8216bc07a089bdd55a4c37cdb426829b6fa4.

packages/perennial-account/contracts/Account.sol
EdNoepel: Done in efc2cf8e1e9c1bb7e828c21f1710fbebed870923, providing a few benefits:
- Now that implementation provides the stablecoin addresses, the initializer only needs to worry about _owner_, as such, we can initialize through `create2` rather than afterward, presumably saving gas.
- _Controller_ no longer needs to know the DSU reserve address, presumably reducing code size.

Regarding that last point, we could go a step further and have `Controller` pull USDC and DSU addresses from the `Account` implementation in _Controller's_ constructor.  Thoughts?

packages/perennial-account/contracts/Controller.sol
EdNoepel: No; I thought I was "losing" the constructor by transitioning creation to a `BeaconProxy`.  Seems much cleaner now that we're baking stablecoin addresses into the implementation at deployment time.

packages/perennial-account/contracts/Account.sol
kbrizzle: Oops yes, I had my comment flipped.

And, yeah I like the idea of pulling these from account so we only have to set them once within the extension.

EdNoepel: Per discussion, done in eac9745d5283c44ab1520ec2db4360aaa16f2559.

kbrizzle (approved):

PR #368, opened 2024-07-04
Resolve collateral accounts integration test issue

EdNoepel:
Collateral accounts had been using forked oracles, intended to decouple the extension from the oracle implementation.  This meant tests would always be using the oracles deployed at a specific address and forked block height.  As such, when the interface between markets and oracles changes, we would only be testing with one side of the change, which is undesirable.

I explored using a Chainlink setup similar to `packages/perennial` integration tests, but decided against it because it was more verbose, heavily utilized mocks, the Chainlink feed registry was unavailable on Arbitrum, and was dissimilar to our production setup.

This change deploys a Pyth oracle setup, which enables collateral account tests to use the same technique as before for commiting prices: impersonate the keeper factory and commit directly to the keeper oracle, exercising the settlement callback.


kbrizzle (approved):

kbrizzle:
🔥

PR #370, opened 2024-07-05
Remove no-op updates

kbrizzle:
- Removes `latestPrice` from `Global` in favor of directly using oracle price now that [all prices are available](https://github.com/equilibria-xyz/perennial-v2/pull/363).
- Removes a significant amount of special-casing in the invariant that supported settling via "no-op updates".

#### Invariant Simplification
 - `MarketInvalidProtectionError`
   - Require collateral change to be zero (could previously deposit, simplifies maintenance check)
   - Allow order position change of zero (removes the need for an additional position check on top of the "closes all" check)
 - `MarketStalePriceError`
   - Require for all updates (removes exception for zero position accounts and deposits)
 - `MarketProtectedError`
   - Require for all updates (removes exception for order position change of zero)
 - `MarketInsufficientCollateralError`
   - Require for all updates (removes exception allowing partial re-collateralization)

#### Migration Notes
- `latestPrice` must be valid prior to upgrading.
  - `Global.latestPrice` will not be read after upgrading, but the Oracle will not have lookback data yet prior to upgrade so we must of the `latestPrice` up-to-date (valid) at the time of upgrading.

### DIFF
--- a/packages/perennial/test/unit/market/Market.test.ts
+++ b/packages/perennial/test/unit/market/Market.test.ts
@@ -369,16 +369,7 @@ const EXPECTED_INTEREST_WITHOUT_FEE_10_67_96_ALL = EXPECTED_INTEREST_10_67_96_AL
 
 async function settle(market: Market, account: SignerWithAddress, sender?: SignerWithAddress) {
   const local = await market.locals(account.address)
-  return await market
-    .connect(sender || account)
-    ['update(address,uint256,uint256,uint256,int256,bool)'](
-      account.address,
-      ethers.constants.MaxUint256,
-      ethers.constants.MaxUint256,
-      ethers.constants.MaxUint256,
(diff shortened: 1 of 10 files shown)


packages/perennial/test/unit/market/Market.test.ts
```diff
@@ -691,7 +682,7 @@ describe('Market', () => {
       })
     })
 
-    describe('#updateRiskParameter', async () => {
+    describe.only('#updateRiskParameter', async () => {
```
EdNoepel, line 685: `.only` was left in

EdNoepel (commented):
comments inline

EdNoepel (commented):
Changes look fine, though some `Global` unit tests are failing; I can reproduce the failure locally.

PR #376, opened 2024-07-20
[Cleanup] Move valid from/to into KeeperOracleParameter

kbrizzle:
- Renames `ProviderParameter` to `KeeperOracleParameter` since these are not generic to all oracle providers.
- Moves `validFrom` `validTo` into the `KeeperOracleParameter` so they can be more easily updated.


packages/perennial-oracle/contracts/interfaces/IKeeperFactory.sol
```diff
@@ -44,8 +44,6 @@ interface IKeeperFactory is IOracleProviderFactory, IFactory, IKept {
     // sig: 0x0afa0593
     error KeeperFactoryVersionOutsideRangeError();
 
-    function validFrom() external view returns (uint256);
```
arjun-io, line 47: do we want to keep the getters (read them from the parameter) to maintain backwards compatibility? not sure if any contracts are actually reading these values...

arjun-io (approved):

packages/perennial-oracle/contracts/interfaces/IKeeperFactory.sol
kbrizzle: I think it's kind of expected that the implementation details of the oracle providers themselves will change a bit from time to time since we have the update pattern.

PR #386, opened 2024-08-02
Decouple settle / sync

kbrizzle:
#### Background
During settle-only mode, new prices can still be committed un-requested causing a new version to cut via `sync()`.

This makes it very hard to ensure all local accounts are synced up to the latest global during an sync migration.

#### Changes
- Splits settle() and sync()
  - `settle()` only settles up to latest order
  - `sync()` settles and then syncs up to the latest timestamp
- Make settle-only mode only allow settle()

#### Migration Notes
While this makes migrations significantly safer, we still require the `makerValue`/`longValue`/`shortValue` to be continually backwards compatible in `Version`.


arjun-io:
I think this is a somewhat dangerous change for integrators since we're keeping the same function signature but changing what it does. My vote would be to keep `settle` the same and introduce a new function for getting up to the latest order. Alternatively when we're in settle only mode maybe we can just not sync in the `settle` call?

PR #391, opened 2024-08-05
Trigger Orders extension

EdNoepel:
Recreates functionality of trigger orders from _MultiInvoker_, but without support for collateral movement.

## TODO
- [x] Get CI working
- [x] Negative tests, particularly for trigger order storage overflows
- [x] Address linter concerns
- [x] Delegated signer tests
- [x] Resolve merge conflicts
- [x] Rename events to disambiguate from _Market_'s "order"


packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+
+import { Fixed6, Fixed6Lib } from "@equilibria/root/number/types/UFixed6.sol";
+import { UFixed6, UFixed6Lib } from "@equilibria/root/number/types/UFixed6.sol";
+import { IMarket, OracleVersion, Order, Position } from "@equilibria/perennial-v2/contracts/interfaces/IMarket.sol";
+
+// TODO: would prefer making side uint8 for logability, and reserving 0 as invalid/empty,
+// but leaving these same as MultiInvoker for backward compatibility
```
EdNoepel, line 9: Per discussion with @kbrizzle today; I will change these enums, breaking backwards compatibility for the purpose of improving the implementation.

packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -0,0 +1,159 @@
...
+}
+
+struct StoredTriggerOrder {
+    /* slot 0 */
+    uint8 side;       // 0 = maker, 1 = long, 2 = short
+    int8 comparison;  // -1 = lte, 1 = gte
+    int64 price;      // <= 9.22t
+    int64 delta;      // <= 9.22t
+    uint64 maxFee;    // < 18.45t
+    address referrer;
```
kbrizzle, line 105: if you add some unallocated padding here above `referrer` you can ensure it cleanly starts on the beginning of slot 1 instead of being spread across 0 and 1.

packages/perennial-order/contracts/Manager.sol
```diff
@@ -0,0 +1,183 @@
...
+    /// @dev Verifies EIP712 messages for this extension
+    IOrderVerifier public verifier;
+
+    /// @dev Stores trigger orders while awaiting their conditions to become true
+    /// Market => User => Nonce => Order
+    mapping(IMarket => mapping(address => mapping(uint256 => TriggerOrderStorage))) private _orders;
+
+    /// @dev Prevents user from reusing orderIds
+    /// User => Nonce => true if spent
+    mapping(address => mapping(uint256 => bool)) private _spentOrderIds;
```
kbrizzle, line 38: Instead of having a totally separate spent mapping, what if we had a single `bool` in the `TriggerOrder` type that was always true unless the value was initialized / deleted. Then on execution / cancel we can just delete the whole trigger order.

packages/perennial-order/contracts/Manager.sol
```diff
@@ -0,0 +1,183 @@
...
+            UFixed6Lib.MAX,
+            Fixed6Lib.from(-1, raisedKeeperFee),
+            false
+        );
+
+        return UFixed18Lib.from(raisedKeeperFee);
+    }
+
+    function _cancelOrder(IMarket market, address user, uint256 orderNonce) private
+    {
```
kbrizzle, line 164: nit: no newline for the bracket (couple of these)

packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -0,0 +1,159 @@
...
+        address user
+    ) internal {
+        // settle and get the pending position of the account
+        market.settle(user);
+        Order memory pending = market.pendings(user);
+        Position memory position = market.positions(user);
+        position.update(pending);
+
+        // apply order to position
+        if (self.side == 0)
```
kbrizzle, line 60: you could maybe put part of this in a helper function so it looked more like:
```
 if (self.side == 0)  position.maker = _add(position.maker, self.delta);
```

packages/perennial-order/test/unit/Manager.test.ts
```diff
@@ -0,0 +1,383 @@
...
+      // submit a replacement with the same order nonce
+      await expect(manager.connect(userA).placeOrder(market.address, nextOrderNonce, replacement))
+        .to.emit(manager, 'OrderPlaced')
+        .withArgs(market.address, userA.address, replacement, nextOrderNonce)
+
+      const order = await manager.orders(market.address, userA.address, nextOrderNonce)
+      compareOrders(order, replacement)
+    })
+
+    it('keeper can execute orders', async () => {
```
kbrizzle, line 159: I'd add a few more cases around execution:
- more sides
- more comparisons
- w/ pending positions on the account

(this is kind of the meat of the important tests for trigger orders)

packages/perennial-order/contracts/Manager.sol
```diff
@@ -0,0 +1,183 @@
...
+        _spentOrderIds[user][orderNonce] = true;
+        emit OrderCancelled(market, user, orderNonce);
+    }
+
+    function _placeOrder(IMarket market, address user, uint256 orderNonce, TriggerOrder calldata order) private
+    {
+        // prevent user from reusing an order identifier
+        if (_spentOrderIds[user][orderNonce]) revert ManagerInvalidOrderNonceError();
+
+        _orders[market][user][orderNonce].store(order);
```
kbrizzle, line 180: Related to the top comment: in your implementation I see that once an order has been spent you can no longer update the slot with a new active order. I think we should allow this, but let me know if you're thinking otherwise.

EdNoepel: I'm looking at `orderNonce` as a client orderId.  If we allowed reuse, we could not distinguish two orders placed in the same slot.  This complicates the cancel/replace workflow, as users might try to replace an order which a keeper just fulfilled.  In this situation, the user's replacement would become an unintentional new order.  

Transitioning from `orderNonce` to `orderSlot` would also bear some complexity to indexers, who would now need to combine the slot with a transaction hash to form a unique identifier.

packages/perennial-order/contracts/types/TriggerOrder.sol
EdNoepel: Done in ece17e5de8b68ed417a173df643e8507db2389d6

packages/perennial-order/contracts/Manager.sol
EdNoepel: Fixed in ece17e5de8b68ed417a173df643e8507db2389d6

packages/perennial-order/contracts/types/TriggerOrder.sol
EdNoepel: My bad; I meant to add a TODO when adding referrer.  Was hoping to squeeze these into a single slot, but types would need to be too small.  Fixed in ece17e5de8b68ed417a173df643e8507db2389d6.

packages/perennial-order/test/unit/Manager.test.ts
EdNoepel: Added unit tests which check all permutations of sides and comparisons through the _Manager_.
Added integration tests for closing positions and changing position with pending orders.

packages/perennial-order/contracts/Manager.sol
kbrizzle: I see, yeah that seems like good extra protection to add. I think my point still holds on packing the bit into the TriggerOrder for the spent flag, does that make sense?

EdNoepel: The original comment was that we should have a flag which was always true, which would become false if exceuted/cancelled (and the order deleted).  If we did this, and eliminated the _spent_ mapping, we could not distinguish between `orderNonces` which were executed/cancelled, and those which were never used.

I could do the opposite, adding a `spent` flag to `TriggerOrder`.  From a DX perspective, we would need to decide whether to revert if user sets  it _true_ upon placement, or to set it false regardless of what user set.  This means instead of deleting executed/cancelled orders, we would change the flag, and could eliminate the _spent_ mapping.  Seems wasteful to leave the whole order onchain, but we no longer receive big gas refunds for deletions.

I can branch off to explore the latter.

packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -0,0 +1,159 @@
...
+        address user
+    ) internal {
+        // settle and get the pending position of the account
+        market.settle(user);
+        Order memory pending = market.pendings(user);
+        Position memory position = market.positions(user);
+        position.update(pending);
+
+        // apply order to position
+        if (self.side == 3) position.maker = _add(position.maker, self.delta);
```
arjun-io, line 59: It looks like we don't handle the `delta` magic value here. The old system allowed for `0` as the "close-all" magic value, but I think this might be unsafe since 0 is the default. Potentially we use min int64 instead?

EdNoepel: `_add` method handles the magic value, and I recently wrote a unit test to exercise it.  Agree that `type(int64).minValue` seems safer; will adjust accordingly.

packages/perennial-order/contracts/Manager.sol
```diff
@@ -0,0 +1,181 @@
...
+            UFixed6Lib.MAX,
+            UFixed6Lib.MAX,
+            Fixed6Lib.from(-1, raisedKeeperFee),
+            false
+        );
+
+        return UFixed18Lib.from(raisedKeeperFee);
+    }
+
+    function _cancelOrder(IMarket market, address user, uint256 orderNonce) private {
```
kbrizzle, line 163: lets change this to `orderId` since we have a convention of calling these mapping indexes ids throughout the protocol.

kbrizzle (approved):

packages/perennial-order/contracts/Manager.sol
EdNoepel: Done in 9bfe1fc75616f1b2831346fbbc2e1a24b589543b.

packages/perennial-order/contracts/types/TriggerOrder.sol
EdNoepel: Done in [02a44e8](https://github.com/equilibria-xyz/perennial-v2/pull/391/commits/02a44e87651dad789c5ec380eaef7be40dc45ce0).

packages/perennial-order/contracts/Manager.sol
EdNoepel: Done in [5ac7554](https://github.com/equilibria-xyz/perennial-v2/pull/391/commits/5ac755404a95382c9c64344f3ed832398f0c1b53).

PR #396, opened 2024-08-07
Add support for operator on claimFee

EdNoepel:
## Changes
- add `account` parameter to `IMarket.claimFee`
- introduce `onlyOperator` modifier
- update `FeeClaimed` event to show both who earned the fee and who claimed it

## TODO
- [x] update extensions which use `claimFee`
- [x] get CI running cleanly
- [x] resolve merge conflicts

### DIFF
--- a/packages/common/testutil/time.ts
+++ b/packages/common/testutil/time.ts
@@ -35,8 +35,9 @@ export async function increaseTo(timestamp: number): Promise<void> {
     await time.increaseTo(timestamp)
   }
   const newTimestamp = await currentBlockTimestamp()
-  if (timestamp != newTimestamp)
-    console.log('[WARNING] increaseTo failed to reach timestamp (%s vs %s)', timestamp, newTimestamp)
+  if (timestamp != newTimestamp) {
+    await setNextBlockTimestamp(timestamp)
+  }
 }
 
 export async function reset(blockNumber?: number): Promise<void> {
--- a/packages/common/testutil/time.ts
+++ b/packages/common/testutil/time.ts
@@ -35,9 +35,8 @@ export async function increaseTo(timestamp: number): Promise<void> {
     await time.increaseTo(timestamp)
   }
   const newTimestamp = await currentBlockTimestamp()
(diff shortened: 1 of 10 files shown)


packages/common/testutil/time.ts
```diff
@@ -35,8 +35,9 @@ export async function increaseTo(timestamp: number): Promise<void> {
     await time.increaseTo(timestamp)
   }
   const newTimestamp = await currentBlockTimestamp()
-  if (timestamp != newTimestamp)
-    console.log('[WARNING] increaseTo failed to reach timestamp (%s vs %s)', timestamp, newTimestamp)
+  if (timestamp != newTimestamp) {
+    await setNextBlockTimestamp(timestamp)
```
EdNoepel, line 39: Experimenting to see if this works any better.  Current perspective is that the test still fails, but this provides a better error, indicating you cannot set the next block's timestamp earlier than the previous block.

EdNoepel:
Note I can locally reproduce the `PythOracleFactory` oracle integration test on `v2.3` branch running `yarn coverage:integration` with an `.only` on that test suite.

packages/perennial-extensions/contracts/Coordinator.sol
```diff
@@ -36,7 +36,7 @@ contract Coordinator is ICoordinator, Ownable {
     /// @param market The market to claim the fee for
     function claimFee(IMarket market) external {
         if (msg.sender != comptroller) revert NotComptroller();
-        market.claimFee();
+        market.claimFee(comptroller);
```
kbrizzle, line 39: We'd want this to be `msg.sender`.

We want the `Coordinator` contract to be the Market's coordinator role, rather than using a delegated claim (because we're mainly solving the authorization issue here).

EdNoepel: Done in 473e1b46228eea62f440aef9ebbbc650543348d2, and adjusted unit test to check the address passed to the call.

arjun-io (approved):

kbrizzle (approved):

PR #397, opened 2024-08-09
Decouple settle / sync

kbrizzle:
#### Background
During settle-only mode, new prices can still be committed un-requested causing a new version to cut via `sync()`.

This makes it very hard to ensure all local accounts are synced up to the latest global during an async migration.

#### Changes
- Splits settle() and sync() / flips naming to ensure backwards compatibility
  - `sync()` only settles up to latest order
  - `settle()` settles up to latest order and then up to the latest timestamp
- Rename settle-only to sync-only, make sync-only mode only allow `sync()`

#### Migration Notes
While this makes migrations significantly safer, we still require the `makerValue`/`longValue`/`shortValue` to be continually backwards compatible in `Version`.

### DIFF
--- a/packages/perennial/test/unit/market/Market.test.ts
+++ b/packages/perennial/test/unit/market/Market.test.ts
@@ -1266,6 +1266,264 @@ describe('Market', () => {
       })
     })
 
+    describe('#sync', async () => {
+      beforeEach(async () => {
+        await market.connect(owner).updateCoordinator(coordinator.address)
+        await market.connect(owner).updateBeneficiary(beneficiary.address)
+        await market.connect(owner).updateParameter(marketParameter)
+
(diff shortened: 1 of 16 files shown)


arjun-io (approved):

packages/perennial/test/unit/market/Market.test.ts
```diff
@@ -1524,6 +1524,177 @@ describe('Market', () => {
...
+          .withArgs(
+            user.address,
+            1,
+            {
+              ...DEFAULT_ORDER,
+              orders: 1,
+              timestamp: ORACLE_VERSION_2.timestamp,
+              collateral: COLLATERAL,
+              makerPos: POSITION,
+            },
```
yang923, line 1600: Looks there are some duplication in the code like above, wondering whether we can extract those into a helper function for reusability?

kbrizzle:
This ends up adding too much code size, so closing in favor of: https://github.com/equilibria-xyz/perennial-v2/pull/398.

PR #402, opened 2024-08-12
Add operator to intent update

kbrizzle:
Adds an `account` parameter to the intent `update()` signifying to the maker in the match. This allows intent orders to be filled by an operator is one is approved for account.


arjun-io (approved):

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -234,6 +238,19 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
+    /// @param account Address of account to update
+    /// @param intent The intent that is being filled
+    /// @param signature The signature of the intent that is being filled
+    function _updateIntent(
+        address account,
+        IMarket market,
+        Intent memory intent,
+        bytes memory signature
+    ) internal isMarketInstance(market) {
+        market.update(account, intent, signature);
```
arjun-io, line 251: Do we want to handle deposits/withdrawals and interfaces fees here as we do in the regular update?

kbrizzle: Because the maker is sending the transaction it's not safe to attach additive interface fees here. And the deposit / withdraw happens separately for intent updates. We may want to discuss how additive fees will work for intents though.

PR #403, opened 2024-08-12
Gas aware liquidation fee

kbrizzle:
Updates the liquidation fee from a fixed dollar amount to a multiple of the settlement fee.

This allows the liquidation fee to automatically scale with the gas cost of the settlement.


packages/perennial/contracts/libs/VersionLib.sol
```diff
@@ -177,7 +177,9 @@ library VersionLib {
         Version memory next,
         VersionAccumulationContext memory context
     ) private pure returns (UFixed6 liquidationFee) {
-        liquidationFee = context.toOracleVersion.valid ? context.riskParameter.liquidationFee : UFixed6Lib.ZERO;
+        liquidationFee = context.toOracleVersion.valid ?
+            context.toOracleReceipt.settlementFee.mul(context.riskParameter.liquidationFee) :
```
arjun-io, line 181: So does this mean the liquidation fee is a multiple of the full settlement fee independent of how many orders requested the oracle version?

kbrizzle: Correct yeah, we're just using the `settlementFee` as a point of reference, but it's not actually tied to the settlement fee flow itself.

arjun-io (approved):

PR #406, opened 2024-08-13
[Code Size] Remove protocol parameter from context

kbrizzle:
Refactors last used field (`referralFee`) of `ProtocolParameter` to return as part of the `authorization` logic instead, which allows us to remove its loading altogether in the `Context`.

Saves `256 bytes` code size.


kbrizzle:
Merging because this change is required to overcome a stack too deep array in [britz-flip-fee-intent](https://github.com/equilibria-xyz/perennial-v2/tree/britz-flip-fee-intent)

PR #408, opened 2024-08-14
Add CLAIM_FEE action to MultiInvoker

EdNoepel:
# Changes
Adds a _MultiInvoker_ action which claims earned market fees for the user and automatically unwraps to USDC.

# TODO
- [x] Merge with v2.3 (notably with PR#402)
- [x] Fix integration test calling from delegate
- [x] Self-review and resolve linter warnings
- [x] Confirm CI runs clean

### DIFF
--- a/packages/perennial-extensions/contracts/MultiInvoker.sol
+++ b/packages/perennial-extensions/contracts/MultiInvoker.sol
@@ -184,6 +184,10 @@ contract MultiInvoker is IMultiInvoker, Kept {
                 (address target) = abi.decode(invocation.args, (address));
 
                 _approve(target);
+            } else if (invocation.action == PerennialAction.CLAIM_FEE) {
+                (IMarket market) = abi.decode(invocation.args, (IMarket));
+
+                _claimFee(account, market);
             }
         }
         // ETH must not remain in this contract at rest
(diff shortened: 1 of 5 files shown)


packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -309,6 +313,14 @@ contract MultiInvoker is IMultiInvoker, Kept {
         emit InterfaceFeeCharged(account, market, interfaceFee);
     }
 
+    /// @notice Claims market fees, unwraps DSU, and pushes USDC to fee earner
+    /// @param market Market from which fees should be claimed
+    /// @param account Address of the user who earned fees
+    function _claimFee(address account, IMarket market) internal isMarketInstance(market) {
```
arjun-io, line 319: We might want to include a `wrap` arg here to keep things in line with how `update` works

EdNoepel: In `_withdraw` method, why is the parameter called `wrap` instead of `unwrap`?  It feels wrong naming this parameter the opposite of what it does.

EdNoepel: Added `unwrap` arg in 6e7ca07e6329f127ec6083549d8d543252737305.

arjun-io: Yeah I'm not opposed to fixing the naming there - I think it's because the `update` arg is named `wrap` when it is specifying both wrapping or unwrapping depending on the direction of movement

kbrizzle (approved):

packages/perennial-extensions/contracts/MultiInvoker.sol
kbrizzle: Yeah, we were using `wrap` to generically mean "wrap or unwrap depending on the case vs do neither" - but I think what you changed it makes more sense.

PR #417, opened 2024-08-19
Update README

Olexandr88:
Update README


EdNoepel (commented):
I'm quite happy with the native Twitter Follow badge.  What is the motivation to wrap it with Shields.io?

Olexandr88:
So I agree, I overdid something

PR #419, opened 2024-08-21
Interface fee for trigger orders

EdNoepel:
Implement an additive interface fee for trigger orders, replete with unwrapping facility.  Fee may be configured as a flat amount or as a percentage of the notional value of the order.

# TODO:
- [x] self-review
- [x] resolve any CI issues


packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -88,27 +91,53 @@ library TriggerOrderLib {
...
-    bytes5 __unallocated__; // padding
+    bytes5 __unallocated0__;      // padding for 32-byte alignment
     /* slot 1 */
     address referrer;
+    bytes12 __unallocated1__;     // padding for 32-byte alignment
+    /* slot 2 */
+    address interfaceFeeReceiver;
+    uint64 interfaceFeeAmount;    // < 18.45t
+    bool interfaceFeeFlat;
+    bool interfaceFeeUnwrap;
```
EdNoepel, line 137: There was plenty of room for these last three in slots 0 and 1, but since we needed slot 2 for the address, figured I would keep all the interface fee stuff together.

packages/perennial-order/package.json
```diff
@@ -12,9 +12,9 @@
     "compile": "hardhat compile",
     "gasReport": "REPORT_GAS=true OPTIMIZER_ENABLED=true yarn test:integration",
     "test": "hardhat test test/unit/*",
-    "test:integration": "FORK_ENABLED=true FORK_NETWORK=arbitrum FORK_BLOCK_NUMBER=208460709 hardhat test test/integration/*",
+    "test:integration": "FORK_ENABLED=true FORK_NETWORK=arbitrum FORK_BLOCK_NUMBER=243648015 hardhat test test/integration/*",
```
EdNoepel, line 15: Needed newer fork so that we're playing with USDC instead of USDCe.

packages/perennial-order/test/integration/Manager_Arbitrum.test.ts
```diff
@@ -645,6 +828,7 @@ describe('Manager_Arbitrum', () => {
         // advance 5 minutes
         await increase(60 * 5)
         advanceBlock()
+        await setNextBlockBaseFee()
```
EdNoepel, line 831: With each iteration of the loop, hardhat's base fee was drifting back towards zero.  This caused the keeper to be undercompensated downstream, trapped by the `afterEach`.

packages/perennial-order/test/integration/Manager_Arbitrum.test.ts
```diff
@@ -663,7 +847,76 @@ describe('Manager_Arbitrum', () => {
       // fulfill oracle version and settle
       await commitPrice(parse6decimal('2000.1'), orderTimestamp)
       await market.settle(userD.address, TX_OVERRIDES)
+      expect((await market.positions(userC.address)).short).to.equal(parse6decimal('2.26'))
       expect((await market.positions(userD.address)).long).to.equal(parse6decimal('3'))
+
+      checkKeeperCompensation = true
+    })
+
+    it('market reverts when attempting to close an unsettled position', async () => {
```
EdNoepel, line 856: Thought it was important to ensure we receive the appropriate revert from the market in this use case, ensuring that the trigger order logic doesn't break and obscure the revert reason.

packages/perennial-order/contracts/Manager.sol
```diff
@@ -132,7 +153,7 @@ abstract contract Manager is IManager, Kept {
         if (user != signer && !marketFactory.signers(user, signer)) revert ManagerInvalidSignerError();
     }
 
-    /// @notice Transfers DSU from market to manager to compensate keeper
+    /// @dev Transfers DSU from market to manager to compensate keeper
```
kbrizzle, line 156: nit: we usually use notice instead of dev for these

packages/perennial-order/contracts/Manager.sol
```diff
@@ -168,6 +182,28 @@ abstract contract Manager is IManager, Kept {
         emit TriggerOrderCancelled(market, user, orderId);
     }
 
+    /// @dev Transfers DSU from market to manager to pay interface fee
+    function _chargeInterfaceFee(IMarket market, address user, TriggerOrder memory order) internal returns (bool) {
```
kbrizzle, line 186: naming: lets stick to `account` for the address throughout here

packages/perennial-order/contracts/Manager.sol
```diff
@@ -176,4 +212,10 @@ abstract contract Manager is IManager, Kept {
         _orders[market][user][orderId].store(order);
         emit TriggerOrderPlaced(market, user, order, orderId);
     }
+
+    /// @dev Unwraps DSU to USDC
+    function _unwrap(address receiver, UFixed18 amount) private {
+        reserve.redeem(amount);
+        if (receiver != address(this)) USDC.push(receiver, UFixed6Lib.from(amount));
```
kbrizzle, line 219: I don't think the if condition is necessary on this one -- the user shouldn't be sending this to `this` and if they do for some reason it would hurt anything.

packages/perennial-order/contracts/Manager.sol
```diff
@@ -176,4 +212,10 @@ abstract contract Manager is IManager, Kept {
         _orders[market][user][orderId].store(order);
         emit TriggerOrderPlaced(market, user, order, orderId);
     }
+
+    /// @dev Unwraps DSU to USDC
+    function _unwrap(address receiver, UFixed18 amount) private {
```
kbrizzle, line 217: Should we name this `_unwrapAndWithdaw`? I was a little confused at first glance why we didn't need to call `.push()` after `_unwrap()`/

packages/perennial-order/contracts/types/InterfaceFee.sol
```diff
@@ -0,0 +1,31 @@
...
+import { UFixed6 } from "@equilibria/root/number/types/UFixed6.sol";
+
+/// @dev Additive fee optionally awarded to GUIs upon execution of trigger orders
+struct InterfaceFee {
+    /// @dev Amount of DSU to transfer from market to recipient
+    UFixed6 amount;
+    /// @dev Recipient of the fee
+    address receiver;
+    /// @dev True if fee is a fixed amount, false if fee is percentage of change in notional value
+    bool flatFee;
```
kbrizzle, line 14: nit: we use "fixed" or "absolute" elsewhere vs "flat" although not a huge deal

EdNoepel: Renamed in 73751c255e1c8571a02792eec1d3f629ba844252.

packages/perennial-order/contracts/Manager.sol
EdNoepel: I had been using `@notice` for user/integrator-facing comments and `@dev` for protocol-developer-facing comments, usually on internal/private methods.  Commit 4c165a197422c15dfbdeda16f0040e614859db41 changes several to _notice_, but feel free to let me know if there's some guidance on how to choose which to use.

packages/perennial-order/contracts/Manager.sol
EdNoepel: Done in 4c165a197422c15dfbdeda16f0040e614859db41.

packages/perennial-order/contracts/Manager.sol
EdNoepel: Removed in 4c165a197422c15dfbdeda16f0040e614859db41.

packages/perennial-order/contracts/Manager.sol
EdNoepel: Agree; done in 4c165a197422c15dfbdeda16f0040e614859db41.

packages/perennial-order/contracts/Manager.sol
kbrizzle: Oh got it, that's different than what we do elsewhere for consistency, but that does make sense, so I'm down to change that standard going forward if we want.

kbrizzle (approved):

PR #423, opened 2024-08-23
Add oracle name

kbrizzle:


packages/perennial-oracle/contracts/Oracle.sol
```diff
@@ -23,6 +23,9 @@ contract Oracle is IOracle, Instance {
     /// @notice The beneficiary of the oracle fee
     address public beneficiary;
 
+    /// @notice The name of the oracle
+    string public name;
+
     /// @notice Initializes the contract state
     /// @param initialProvider The initial oracle provider
     function initialize(IOracleProvider initialProvider) external initializer(1) {
```
arjun-io, line 31: Should we init with a name? Or just update separately each time?

arjun-io (approved):
One question about adding name to the initializer

arjun-io (approved):

PR #431, opened 2024-08-27
add nonce offset support to market creation task

arjun-io:
Quick one to update the hardhat task which builds a multisig payload to launch a new market. When launching multiple markets at once, we need to add the ability to specify a nonce offset when determining the resulting addresses from the factory creation methods


kbrizzle (approved):

kbrizzle:
If the `nonceOffset` is incorrectly set, it will just cause the resulting tx to revert right? Wondering if we need to add any sort of additional fool proofing to this.

arjun-io:
> If the `nonceOffset` is incorrectly set, it will just cause the resulting tx to revert right? Wondering if we need to add any sort of additional fool proofing to this.

Correct, it should because there are checks at different points which assert the sub oracle and oracle exist

PR #444, opened 2024-09-20
Fix verify 

prateekdefi:
Issue: https://github.com/sherlock-audit/2024-08-perennial-v2-update-3-judging/issues/54


kbrizzle:
Since this affects `Verifier`, `AccountVerifier`, and `OrderVerifier` I think it would be better to find a solution that updates the core `VerifierBase` in `root`.

We can:
- Add an internal abstract function in `VerifierBase` `authorized` that defaults to reverting unless `signer == account`.
  - Add the `authorized` check to `validateAndCancel`
- In each of the three `Verifier` implementations, we can add `MarketFactory` as an immutable value (no need to add it as a parameter).
  - We can then override `authorized` as `authorized.super() or signer is delegated`
- As a cleanup step, we can remove the `account is signer` checks from all upstream sources

Caveat:
- The verifications in `MarketFactory` are meant to only allow the account itself (and not its delegated signer) to perform the updates. It's not the cleanest, but we can leave the checks in here as-is to get the desired effect.
- Intents have their own signer verification flow deep in the invariant. I don't think we should remove this even though it will be applying the verification check essentially twice. We should make sure there is a test case to cover submitting an intent order signed by the taker's delegated signing account to ensure it passes both checks.

EdNoepel:
Resolved merge conflicts, though there is some issue with keeper compensation for relayed messages.  Currently investigating.

EdNoepel (approved):
LGTM

packages/perennial-account/contracts/AccountVerifier.sol
```diff
@@ -148,4 +154,14 @@ contract AccountVerifier is VerifierBase, IAccountVerifier {
...
         )) revert VerifierInvalidSignerError();
     }
+
+    /// @notice Checks account authorization
+    /// @param account the account to check authorization for
+    /// @param signer the signer of the account
+    /// @return whether signer is authorized
+    function _authorized(address account, address signer) internal view override returns (bool) {
+        // check if given signer is approved as signer or operator
+        (bool isOperator, bool isSigner, ) = marketFactory.authorization(account, signer, signer, address(0));
```
kbrizzle, line 164: we only want to allow delegated (or self) signers here, not approved operators / extensions, so it should look something like this:

`return super._authorized() || marketFactory.signers(owner, signer))`
(order is important so that self-signs use less gas due to early termination)

for reference, it should essentially match the old logic here: https://github.com/equilibria-xyz/perennial-v2/pull/444/files#diff-7253e5944c084684a615a72fa46c6ec49b4457786f8584fd810fa4966e53eb57L186.

packages/perennial-account/contracts/Controller_Arbitrum.sol
```diff
@@ -4,18 +4,20 @@ pragma solidity 0.8.24;
...
 /// @title Controller_Arbitrum
 /// @notice Controller which compensates keepers for handling or relaying messages on Arbitrum L2.
 contract Controller_Arbitrum is Controller_Incentivized, Kept_Arbitrum {
     /// @dev Creates instance of Controller which compensates keepers
     /// @param implementation Pristine collateral account contract
+    /// @param marketFactory Market Factory contract
     /// @param nonceManager Verifier contract to which nonce and group cancellations are relayed
     constructor(
-        address implementation,
+        address implementation, IMarketFactory marketFactory,
```
kbrizzle, line 18: nit: newline the `marketFactory` param

packages/perennial-account/contracts/Controller_Incentivized.sol
```diff
@@ -46,23 +46,22 @@ abstract contract Controller_Incentivized is Controller, IRelayer, Kept {
 
     /// @dev Creates instance of Controller which compensates keepers
     /// @param implementation_ Pristine collateral account contract
+    /// @param marketFactory_ Market factory contract
     /// @param nonceManager_ Verifier contract to which nonce and group cancellations are relayed
     constructor(
-        address implementation_,
+        address implementation_, IMarketFactory marketFactory_,
```
kbrizzle, line 52: nit: newline the marketFactory param

packages/perennial-order/contracts/OrderVerifier.sol
```diff
@@ -45,4 +51,14 @@ contract OrderVerifier is VerifierBase, IOrderVerifier {
...
         )) revert VerifierInvalidSignerError();
     }
+
+    /// @notice Checks account authorization
+    /// @param account the account to check authorization for
+    /// @param signer the signer of the account
+    /// @return whether signer is authorized
+    function _authorized(address account, address signer) internal view override returns (bool) {
+        // check if given signer is approved as signer or operator
+        (bool isOperator, bool isSigner, ) = marketFactory.authorization(account, signer, signer, address(0));
```
kbrizzle, line 61: same note about `_authorized` logic here.

packages/perennial/test/integration/core/happyPath.test.ts
```diff
@@ -1608,6 +1606,9 @@ describe('Happy Path', () => {
     intent.solver = userB.address
     signature = await signIntent(userC, verifier, intent)
 
+    // userC is allowed to sign messages for user
+    await marketFactory.connect(user).updateSigner(userC.address, true)
```
kbrizzle, line 1610: we shouldn't have to change the authorization logic in any of these tests, this might indicate a bug in the implementation.

packages/perennial/test/integration/core/happyPath.test.ts
```diff
@@ -1690,15 +1689,13 @@ describe('Happy Path', () => {
         account: user.address,
         signer: userC.address,
         domain: market.address,
-        nonce: 0,
+        nonce: 1,
```
kbrizzle, line 1692: why did we need to change this?

packages/perennial-verifier/contracts/interfaces/IMarketFactory.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+import "@equilibria/root/number/types/UFixed6.sol";
+
+interface IMarketFactory {
```
kbrizzle, line 5: can we import the real one instead of add this?

packages/perennial-verifier/contracts/interfaces/IVerifier.sol
```diff
@@ -9,6 +9,8 @@ import { SignerUpdate } from "../types/SignerUpdate.sol";
 import { AccessUpdateBatch } from "../types/AccessUpdateBatch.sol";
 
 interface IVerifier is IVerifierBase {
+    error VerifierMarketFactoryZeroAddressError();
```
kbrizzle, line 12: i think this is unused

packages/perennial-verifier/contracts/Verifier.sol
```diff
@@ -88,4 +99,14 @@ contract Verifier is VerifierBase, IVerifier {
...
         )) revert VerifierInvalidSignerError();
     }
+
+    /// @notice Checks account authorization
+    /// @param account the account to check authorization for
+    /// @param signer the signer of the account
+    /// @return whether signer is authorized
+    function _authorized(address account, address signer) internal view override returns (bool) {
+        // check if given signer is approved as signer or operator
+        (bool isOperator, bool isSigner, ) = marketFactory.authorization(account, signer, signer, address(0));
```
kbrizzle, line 109: same note about _authorized logic here.

packages/perennial-verifier/test/unit/Verifier/Verifier.test.ts
```diff
@@ -55,6 +66,28 @@ describe('Verifier', () => {
...
+        account: caller.address,
+        signer: caller2.address,
+        domain: caller2.address,
+        nonce: 0,
+        group: 0,
+        expiry: constants.MaxUint256,
+      }
+      const signature = await signCommon(caller2, verifier, commonMessage)
+
+      marketFactory.authorization.returns([false, false, BigNumber.from(0)])
```
kbrizzle, line 81: will need to be updated with new `_authorized` logic.

packages/perennial-verifier/test/unit/Verifier/Verifier.test.ts
```diff
@@ -55,6 +66,28 @@ describe('Verifier', () => {
...
+
+      await expect(verifier.connect(caller2).verifyCommon(commonMessage, signature)).to.be.revertedWithCustomError(
+        verifier,
+        'VerifierInvalidSignerError',
+      )
+
+      expect(await verifier.nonces(caller.address, 0)).to.eq(false)
+      marketFactory.authorization.reset()
```
kbrizzle, line 89: i dont think you need to reset at the end here

packages/perennial-verifier/test/unit/Verifier/Verifier.test.ts
```diff
@@ -55,6 +66,28 @@ describe('Verifier', () => {
 
       expect(await verifier.nonces(caller.address, 0)).to.eq(true)
     })
+
+    it('should reject common w/ invalid signer or operator', async () => {
```
kbrizzle, line 70: we should probably have a test like this in each of the three verifier implementations

packages/perennial/test/integration/core/happyPath.test.ts
EdNoepel: The intent acts on account `user` but is signed by `userC`: see line 1529.  Seems authorization logic should have been needed before.  Perhaps the old implementation was producing the expected revert before validating the signer.

packages/perennial-account/contracts/AccountVerifier.sol
EdNoepel: done in 4e960a48d420e86fae7492e52a5916663480840b

packages/perennial-account/contracts/Controller_Arbitrum.sol
EdNoepel: done in 4e960a48d420e86fae7492e52a5916663480840b

packages/perennial-order/contracts/OrderVerifier.sol
EdNoepel: done in 4e960a48d420e86fae7492e52a5916663480840b

packages/perennial-account/contracts/Controller_Incentivized.sol
EdNoepel: done in 4e960a48d420e86fae7492e52a5916663480840b

packages/perennial-verifier/contracts/interfaces/IMarketFactory.sol
EdNoepel: `IMarketFactory` imports `IVerifier` from here, which would create a circular dependency.  Some options:
A) **Define `IVerifier` in core package, and have verifier package import it from there** - This seems more consistent with how we do things.  For example, the oracle package pulls _OracleVersion_ from core.
B) **Reimplement _IVerifier_ in core package**, moving the problem there.
C) **Leave as is** - Undesirable, because _IMarketFactory_ is more likely to change, increasing maintenance cost.

I'll go with A for now.

packages/perennial-verifier/contracts/interfaces/IVerifier.sol
EdNoepel: removed in 9e58893a8b8318ce2fa41d924c7cf03f9f4d60ed

packages/perennial-verifier/contracts/Verifier.sol
EdNoepel: changed logic in 91d3cd09b871fdb72a820d1f34f8973ad01351d4

packages/perennial-verifier/test/unit/Verifier/Verifier.test.ts
EdNoepel: removed in 91d3cd09b871fdb72a820d1f34f8973ad01351d4

packages/perennial-verifier/test/unit/Verifier/Verifier.test.ts
EdNoepel: added in d13892dc240a6b6fca371f64706dc48307b862a7

packages/perennial/test/integration/core/happyPath.test.ts
EdNoepel: On line 1651, we use nonce 0 to delegate _userC_ as signer for _user_.  The previous implementation cancelled this nonce on a different verifier instance, allowing it to be reused.

EdNoepel:
Extensions unit test failure came from fix reveiew branch; will deal with that outside of this PR.

packages/perennial-verifier/contracts/interfaces/IMarketFactory.sol
kbrizzle: Ahh I see, let's leave as is then. We might want to pull in this Verifier into the `perennial` package now that we have multiple Verifier anyways in v2.4.

packages/perennial-account/test/unit/Controller.ts
```diff
@@ -84,11 +84,17 @@ describe('Controller', () => {
...
-    controller = await deployController(owner, usdc.address, dsu.address, reserve.address)
-
     marketFactory = await smock.fake<IMarketFactory>('IMarketFactory')
-    verifier = await new AccountVerifier__factory(owner).deploy()
-    await controller.initialize(marketFactory.address, verifier.address)
+
+    controller = await deployController(owner, usdc.address, dsu.address, reserve.address, marketFactory.address)
+    verifier = await new AccountVerifier__factory(owner).deploy(marketFactory.address)
+    await controller.initialize(verifier.address)
+    marketFactory.authorization
```
kbrizzle, line 92: are these two mocks still needed with the change to `marketFactory.signers()`?

packages/perennial/test/integration/core/happyPath.test.ts
kbrizzle: Ahh yeah that's it, it reverting earlier now that the signer check is in the beginning as well as the invariant.

kbrizzle (approved):

packages/perennial-account/test/unit/Controller.ts
EdNoepel: nope; removed in 43df24e1ec012811a0d55d40f1cf0d5c159f4c38

PR #451, opened 2024-09-25
Fix issue with keeper compensation

EdNoepel:
Fixes https://github.com/sherlock-audit/2024-08-perennial-v2-update-3-judging/issues/92

### DIFF
--- a/packages/perennial-account/contracts/Controller_Incentivized.sol
+++ b/packages/perennial-account/contracts/Controller_Incentivized.sol
@@ -28,6 +28,8 @@ import { RelayedSignerUpdate } from "./types/RelayedSignerUpdate.sol";
 import { RelayedAccessUpdateBatch } from "./types/RelayedAccessUpdateBatch.sol";
 import { Withdrawal } from "./types/Withdrawal.sol";
 
+import "hardhat/console.sol";
+
 /// @title Controller_Incentivized
 /// @notice Controller which compensates keepers for handling or relaying messages. Subclass to handle differences in
 /// gas calculations on different chains.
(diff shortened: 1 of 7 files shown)


packages/perennial-account/contracts/Controller_Incentivized.sol
```diff
@@ -35,77 +35,116 @@ abstract contract Controller_Incentivized is Controller, IRelayer, Kept {
...
-        // if we're depositing collateral to the market, pay the keeper before transferring funds
-        if (marketTransfer.amount.gte(Fixed6Lib.ZERO)) {
-            _handleKeeperFee(keepConfig, 0, msg.data[0:0], 0, data);
-            _marketTransferWithSignature(account, marketTransfer, signature);
-        // otherwise handle the keeper fee normally, after withdrawing to the collateral account
-        } else {
-            _marketTransferWithSignature(account, marketTransfer, signature);
-            _handleKeeperFee(keepConfig, 0, msg.data[0:0], 0, data);
-        }
+        _marketTransferWithSignature(account, marketTransfer, signature);
```
kbrizzle, line 130: how are we handling deposits now that we always charge the fee afterwards?

kbrizzle:
`constructor` / `initializer()` LGTM -- will let @arjun-io verify the incentivization setup.

packages/perennial-account/contracts/Controller_Incentivized.sol
EdNoepel: A "full deposit" is a corner case because fees are always paid from the account.  If a user was able to transfer their entire collateral account into one or more markets, they could not pay a keeper to rebalance.  When we last discussed this, we concluded an accurate gas calculation is worth the UX tradeoff.

PR #455, opened 2024-10-01
Account for pending position in notional value calculation

EdNoepel:
Fixes https://github.com/sherlock-audit/2024-08-perennial-v2-update-3-judging/issues/62

### DIFF
--- a/packages/perennial-order/contracts/Manager.sol
+++ b/packages/perennial-order/contracts/Manager.sol
@@ -126,8 +126,8 @@ abstract contract Manager is IManager, Kept {
         if (!canExecute) revert ManagerCannotExecuteError();
 
         _compensateKeeper(market, account, order.maxFee);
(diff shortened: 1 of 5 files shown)


packages/perennial-order/contracts/Manager.sol
```diff
@@ -126,8 +126,8 @@ abstract contract Manager is IManager, Kept {
         if (!canExecute) revert ManagerCannotExecuteError();
 
         _compensateKeeper(market, account, order.maxFee);
-        order.execute(market, account);
         bool interfaceFeeCharged = _chargeInterfaceFee(market, account, order);
+        order.execute(market, account);
```
EdNoepel, line 130: Interface fee needs to be charged before (another) pending order is introduced.

kbrizzle (approved):

PR #456, opened 2024-10-01
Fix issue with settle() asyncFee is left in the KeepFactory and is not transfer to the keeper

prateekdefi:
Issue: https://github.com/sherlock-audit/2024-08-perennial-v2-update-3-judging/issues/34


prateekdefi:
@EdNoepel  @kbrizzle  @arjun-io. We can also do this by passing the keeper address from keeper factory to keeper oracle. That would be more gas efficient(only a single push) but it would change the logic for both the factory and the oracle. What do u guys think is a better option.

EdNoepel (commented):
Perhaps we should add a test which settles and checks keeper balance to ensure they got the fee?  Such a test should have failed with the old implementation.

packages/perennial-oracle/contracts/keeper/KeeperFactory.sol
```diff
@@ -164,8 +164,10 @@ abstract contract KeeperFactory is IKeeperFactory, Factory {
         if (oracleIds.length == 0 || oracleIds.length != versions.length || oracleIds.length != maxCounts.length)
             revert KeeperFactoryInvalidSettleError();
 
-        for (uint256 i; i < oracleIds.length; i++)
-            IKeeperOracle(address(oracles[oracleIds[i]])).settle(versions[i], maxCounts[i]);
+        for (uint256 i; i < oracleIds.length; i++) {
+            Token18 rewardToken = IKeeperOracle(address(oracles[oracleIds[i]])).oracle().market().token();
```
kbrizzle, line 168: i think this line is unused, we just need to pass the new receiver in

kbrizzle:
As @EdNoepel mentioned, would be good to have an additional test that fails on the old implementation.

packages/perennial-oracle/contracts/keeper/KeeperFactory.sol
EdNoepel: removed in f55c199ad350bf0f8f08e316444c298a68aff19a

EdNoepel:
> As @EdNoepel mentioned, would be good to have an additional test that fails on the old implementation.

Upon review, this is exactly what `settles successfully and incentivizes the keeper` was testing.  Just the amount was wrong.  Attempted to calculate it, but the fee depends on gas consumption and such.

arjun-io (approved):

PR #458, opened 2024-10-01
Fix USDC unwrapping

prateekdefi:
Issue: https://github.com/sherlock-audit/2024-08-perennial-v2-update-3-judging/issues/35

### DIFF
--- a/packages/perennial-extensions/contracts/MultiInvoker.sol
+++ b/packages/perennial-extensions/contracts/MultiInvoker.sol
@@ -367,8 +367,9 @@ contract MultiInvoker is IMultiInvoker, Kept {
     function _unwrap(address receiver, UFixed18 amount) internal {
         // If the batcher is 0 or doesn't have enough for this unwrap, go directly to the reserve
         if (address(batcher) == address(0) || amount.gt(UFixed18Lib.from(USDC.balanceOf(address(batcher))))) {
+            UFixed6 balanceBefore = USDC.balanceOf(address(this));
             reserve.redeem(amount);
-            if (receiver != address(this)) USDC.push(receiver, UFixed6Lib.from(amount));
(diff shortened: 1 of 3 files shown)


EdNoepel (commented):
Seems we should have a unit test which exercises the scenario explained in the report: The call to `reserve.redeem(amount)` returns less than `amount`, so the difference in USDC balance != `amount`.

packages/perennial-extensions/contracts/MultiInvoker.sol
```diff
@@ -367,8 +367,9 @@ contract MultiInvoker is IMultiInvoker, Kept {
     function _unwrap(address receiver, UFixed18 amount) internal {
         // If the batcher is 0 or doesn't have enough for this unwrap, go directly to the reserve
         if (address(batcher) == address(0) || amount.gt(UFixed18Lib.from(USDC.balanceOf(address(batcher))))) {
+            UFixed6 balanceBefore = USDC.balanceOf(address(this));
             reserve.redeem(amount);
-            if (receiver != address(this)) USDC.push(receiver, UFixed6Lib.from(amount));
+            if (receiver != address(this)) USDC.push(receiver, USDC.balanceOf(address(this)).sub(balanceBefore));
```
arjun-io, line 372: I think we can just pass the return value of `redeem` into the push: https://github.com/equilibria-xyz/emptyset-mono/blob/britz-interest-bearing-reserve/packages/emptyset-reserve/contracts/interfaces/IReserve.sol#L51

arjun-io: Actually this would require us to upgrade the Reserve before migrating to v2.3. We can just add this as an optimization to v2.4

arjun-io (approved):

kbrizzle (approved):

PR #459, opened 2024-10-02
Fix liquidation fee risk parameter

prateekdefi:
Issue: https://github.com/sherlock-audit/2024-08-perennial-v2-update-3-judging/issues/69

### Migration notes
Due to changes in `RiskParameter` and `ProtocolParameter` storage layouts, Individual risk parameters and protocol parameters must be updated atomically on upgrade.


EdNoepel (commented):
The point of the report was that the `liquidationFee` risk parameter is a percentage, yet it was being validated against fixed values.  One of those validations has been removed.  The other validation was moved to compare it with a new `maxLiquidationFeeMultiplier` protocol parameter.  The natspec does not state _maxLiquidationFeeMultiplier_ is a percentage.  The term "multiplier" in the name suggests it, though I find the naming confusing.

A majority of tests are setting it to 1000%, which doesn't seem realistic.

packages/perennial/contracts/types/ProtocolParameter.sol
```diff
@@ -8,8 +8,8 @@ struct ProtocolParameter {
     /// @dev The maximum for market fee parameters
     UFixed6 maxFee;
 
-    /// @dev The maximum for market absolute fee parameters
-    UFixed6 maxFeeAbsolute;
+    /// @dev The maximum for liquidation fee parameters
+    UFixed6 maxLiquidationFeeMultiplier;
```
kbrizzle, line 12: let's call this `maxLiquidationFee`, we are simply capping the `liquidationFee` risk parameter, so there's no reason to introduce extra confusion around whether it's a multiplier or not.

packages/perennial/contracts/types/ProtocolParameter.sol
```diff
@@ -32,7 +32,7 @@ struct ProtocolParameter {
 struct StoredProtocolParameter {
     /* slot 0 (28) */
     uint24 maxFee;                  // <= 1677%
-    uint48 maxFeeAbsolute;          // <= 281m
+    uint48 maxLiquidationFeeMultiplier;          // <= 281m
```
kbrizzle, line 35: `uint32` is probably good for this since it's a multiple.

since we reset the parameters on upgrade anyways, reshuffling the layout is ok, let's just make a note of it.

kbrizzle:
Agree with @EdNoepel as well on the value for `maxLiquidationFee` throughout the tests. It should generally be something like double what we're actually using for `liquidationFee`, which I believe is `10` in most places, so maybe `20`?

packages/perennial/contracts/types/ProtocolParameter.sol
EdNoepel: Why `uint32` instead of `uint24` for this percentage?  24 bits lets us set it up to 1677%, as with _maxFee_.  I can't envision a need to set this above 100%.

Assume we should reduce `RiskParameter.liquidationFee` to same size.

EdNoepel: I now understand that, contrary to natspec, that `liquidationFee` is a multiple of the settlement fee.  Will correct the natspec, which states otherwise.

kbrizzle: Yeah, normally we'd implement a multiplier as dividing an inverse so that we can pack it `<100%`, but we did a straight multiplication here so we need to give it breathing room. Might be something to fix in the next version.

packages/perennial/contracts/types/RiskParameter.sol
```diff
@@ -77,12 +77,12 @@ using RiskParameterStorageLib for RiskParameterStorage global;
...
 ///        int32 pControllerMin;                       // <= 214748%
 ///
-///        /* slot 2 */ (32)
+///        /* slot 2 */ (30)
 ///        uint48 pControllerK;                        // <= 281m
 ///        int32 pControllerMax;                       // <= 214748%
 ///        uint48 minMargin;                           // <= 281m
 ///        uint48 minMaintenance;                      // <= 281m
-///        uint48 liquidationFee;                      // <= 281m
+///        uint32 liquidationFee;                      // <= 4294
```
kbrizzle, line 85: Let's add an explicit migration note in the description on this one that the protocol and individual risk parameters must be updated atomically on upgrade.

packages/perennial/test/unit/types/MarketParameter.test.ts
```diff
@@ -32,7 +32,7 @@ export const VALID_MARKET_PARAMETER: MarketParameterStruct = {
 const PROTOCOL_PARAMETER: ProtocolParameterStruct = {
   protocolFee: 0,
   maxFee: parse6decimal('1'),
-  maxFeeAbsolute: BigNumber.from(2).pow(48).sub(1),
+  maxLiquidationFee: BigNumber.from(2).pow(48).sub(1),
```
kbrizzle, line 35: nit: should be `pow(32)` but won't break anything since it's already in memory

packages/perennial/test/unit/types/RiskParameter.test.ts
```diff
@@ -52,9 +52,8 @@ export const VALID_RISK_PARAMETER: RiskParameterStruct = {
 }
 
 const PROTOCOL_PARAMETER: ProtocolParameterStruct = {
-  protocolFee: 0,
   maxFee: parse6decimal('1'),
-  maxFeeAbsolute: BigNumber.from(2).pow(48).sub(1),
+  maxLiquidationFee: BigNumber.from(2).pow(48).sub(1),
```
kbrizzle, line 56: nit: same here for `pow(32)`

kbrizzle (approved):

PR #468, opened 2024-10-10
[Fix] Add maximum for intent price deviation

kbrizzle:
Adds a new market parameter: `maxPriceDeviation` which caps the geometric distance that the quoted intent price can be from the underlying oracle price.

Suggested change from: https://github.com/sherlock-audit/2024-08-perennial-v2-update-3-judging/issues/42.

### DIFF
--- a/packages/perennial/contracts/types/Guarantee.sol
+++ b/packages/perennial/contracts/types/Guarantee.sol
@@ -94,6 +94,19 @@ library GuaranteeLib {
         return self.taker().mul(price).sub(self.notional);
     }
 
+    /// @notice Returns the price deviation of the guarantee from the oracle price
+    /// @dev The price deviation is (intent price - oracle price) / min(intent price, oracle price)
+    ///      Only supports new guarantees for updates, does not work for aggregated guarantees (local / global)
+    /// @param self The guarantee object to check
+    /// @param price The oracle price to compare
+    /// @return The price deviation of the guarantee from the oracle price
+    function priceDeviation(Guarantee memory self, Fixed6 price) internal pure returns (UFixed6) {
+        if (takerTotal(self).isZero()) return UFixed6Lib.ZERO;
+
+        Fixed6 guarenteePrice = self.notional.div(taker(self));
+        return guarenteePrice.sub(price).div(guarenteePrice.min(price)).abs();
+    }
+
     /// @notice Updates the current global guarantee with a new local guarantee
     /// @param self The guarantee object to update
     /// @param guarantee The new guarantee
--- a/packages/perennial/contracts/types/Guarantee.sol
+++ b/packages/perennial/contracts/types/Guarantee.sol
@@ -95,7 +95,7 @@ library GuaranteeLib {
     }
 
     /// @notice Returns the price deviation of the guarantee from the oracle price
-    /// @dev The price deviation is (intent price - oracle price) / min(intent price, oracle price)
+    /// @dev The price deviation is the difference between the prices over the closest price to zero
     ///      Only supports new guarantees for updates, does not work for aggregated guarantees (local / global)
     /// @param self The guarantee object to check
     /// @param price The oracle price to compare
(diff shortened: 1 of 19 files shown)


packages/perennial/contracts/types/Guarantee.sol
```diff
@@ -94,6 +94,19 @@ library GuaranteeLib {
...
     }
 
+    /// @notice Returns the price deviation of the guarantee from the oracle price
+    /// @dev The price deviation is (intent price - oracle price) / min(intent price, oracle price)
+    ///      Only supports new guarantees for updates, does not work for aggregated guarantees (local / global)
+    /// @param self The guarantee object to check
+    /// @param price The oracle price to compare
+    /// @return The price deviation of the guarantee from the oracle price
+    function priceDeviation(Guarantee memory self, Fixed6 price) internal pure returns (UFixed6) {
+        if (takerTotal(self).isZero()) return UFixed6Lib.ZERO;
```
arjun-io, line 104: Should we use `taker(self)` here to correspond with the below `div`?

packages/perennial/contracts/types/Guarantee.sol
```diff
@@ -94,6 +94,19 @@ library GuaranteeLib {
...
+
+        Fixed6 guarenteePrice = self.notional.div(taker(self));
+        return guarenteePrice.sub(price).div(guarenteePrice.min(price)).abs();
```
arjun-io, line 107: Is it safe to assume that `price` and `guaranteePrice` will not be 0 here?

arjun-io (approved):

packages/perennial/contracts/types/Guarantee.sol
kbrizzle: It should not be possible for `price` to be zero due to the latest oracle override.

The `guaranteePrice` could be zero, but this would correctly revert due to being infinitely out of range.

Any reverts here would only block a fresh update, it wouldn't brick the settlement flywheel.

packages/perennial/contracts/types/Guarantee.sol
kbrizzle: `takerTotal` is more correct for measuring whether the guarantee is empty, since you can have an aggregated guarantee net to zero, but we need to use `taker` in the math below for the sign.

packages/perennial/contracts/types/Guarantee.sol
EdNoepel: Might be worth adding a comment stating that, to avoid audit reports on the same.

packages/perennial/contracts/types/MarketParameter.sol
```diff
@@ -49,7 +52,8 @@ using MarketParameterStorageLib for MarketParameterStorage global;
 ///        uint24 riskFee;             // <= 1677%
 ///        uint16 maxPendingGlobal;    // <= 65k
 ///        uint16 maxPendingLocal;     // <= 65k
-///        uint48 __unallocated__;     // <= 281m
+///        uint24 maxPriceDeviation;   // <= 16.77x
```
EdNoepel, line 55: nit: `// <= 1677%` seems more consistent and appropriate here

packages/perennial/contracts/types/Guarantee.sol
```diff
@@ -94,6 +94,19 @@ library GuaranteeLib {
...
+    /// @notice Returns the price deviation of the guarantee from the oracle price
+    /// @dev The price deviation is (intent price - oracle price) / min(intent price, oracle price)
+    ///      Only supports new guarantees for updates, does not work for aggregated guarantees (local / global)
+    /// @param self The guarantee object to check
+    /// @param price The oracle price to compare
+    /// @return The price deviation of the guarantee from the oracle price
+    function priceDeviation(Guarantee memory self, Fixed6 price) internal pure returns (UFixed6) {
+        if (takerTotal(self).isZero()) return UFixed6Lib.ZERO;
+
+        Fixed6 guarenteePrice = self.notional.div(taker(self));
```
EdNoepel, line 106: nit: spelling `guaranteePrice`

EdNoepel (commented):
Some comments inline.

arjun-io (approved):

EdNoepel (approved):
👨‍🍳👌

PR #472, opened 2024-10-15
Cleanup duplicate MarketFactory import

arjun-io:
Having the multiple market factories was causing some issues in the perennial-deploy package's tasks and tests. This introduces a circular dependency but I think both NPM and Solidity should be able to resolve it


arjun-io:
Actually it looks like this doesn't work. I'll just opt to rename the interface in the verifier package

PR #473, opened 2024-10-15
Rename Verifier's market factory interface

arjun-io:
Re-attempt from #472 - Having the multiple market factories was causing some issues in the perennial-deploy package's tasks and tests.


kbrizzle (approved):

kbrizzle:
oof that's annoying

PR #479, opened 2024-10-18
Prateek/pe 1866 code size gas report

prateekdefi:


.github/workflows/gas_report_and_code_size.yml
```diff
@@ -0,0 +1,70 @@
+name: Gas Report and Code Size
```
arjun-io, line 1: One thing that's nice about these tools is that if there is a previous report it'll show the diff as well. Is there any way we could get that working? We might need to commit the reports to the repo

arjun-io: Oh  I see that the code size one does already have the diff

kbrizzle: Yeah gas diff would be incredible

prateekdefi: The tool shows the difference from the last run only. I looked but couldn't find a way to get diff from a previous report. @arjun-io

prateekdefi: Will need to play around in the gas report script to get a diff. Will try that tomorrow.

EdNoepel: Seems the goal here is to pull the "previous" report from the merge target branch.  That implies reports for each branch need to be stored somewhere.  Would also be nice to have this for coverage, though out-of-scope for this PR.

As a reviewer, this information should reveal how much larger _the PR_ is making the code, and how much additional gas will be consumed by transactions impacted by the change.

prateekdefi: Right, the goal is to compare the code size changes and gas changes from base branch. Using a script I tried to compare code size and gas report tables among two branches and show the diff in a single table but that is causing a lot of issues so I think I should not spend more time on this.

I am thinking of running gas report on both the branches in the CI itself and printing two tables one below the other in a single github comment.

prateekdefi:
Created #488  on top of this PR to show how the code size and gas diffs. Could not show the diff on this PR as there is no script added to calculate gas report and code size on v2.3 yet. First I plan to merge this PR to v2.3 and after this #488.

EdNoepel (approved):
The big change here is bumping `hardhat-gas-reporter` to 2.2.1 on base branch, such that #488 can be evaluated properly.

PR #482, opened 2024-10-22
Code formatting with forge

prateekdefi:


EdNoepel (changes requested):
Please update the README with `forge` installation instructions.

packages/perennial-extensions/contracts/interfaces/IMultiInvoker.sol
```diff
@@ -1,27 +1,28 @@
...
+        EXEC_ORDER, // 5
+        COMMIT_PRICE, // 6
         __LIQUIDATE__DEPRECATED,
-        APPROVE,         // 8
-        UPDATE_INTENT,   // 9
-        CLAIM_FEE        // 10
+        APPROVE, // 8
+        UPDATE_INTENT, // 9
+        CLAIM_FEE // 10
+
```
EdNoepel, line 25: This is a good example of something which cannot be maintained when using an auto-formatting tool.  Is implementing an auto-format solution worth the tradeoff of forfeiting our ability to align comments and parameters for readability?

prateekdefi: I agree that auto formatting is not as precise as manual formatting but I think auto formatting is better for following reasons:
- manual formatting is time consuming and human error prone. Adds consistency throughout code.
- code comments help with readability and it would be better if the comments are aligned but it's not that important and can be sacrificed as per me.

packages/perennial/contracts/types/RiskParameter.sol
```diff
@@ -94,55 +88,53 @@ library RiskParameterStorageLib {
     function read(RiskParameterStorage storage self) internal view returns (RiskParameter memory) {
         (uint256 slot0, uint256 slot1, uint256 slot2) = (self.slot0, self.slot1, self.slot2);
         return RiskParameter(
-            UFixed6.wrap(uint256(       slot0 << (256 - 24)) >> (256 - 24)),
-            UFixed6.wrap(uint256(       slot0 << (256 - 24 - 24)) >> (256 - 24)),
+            UFixed6.wrap(uint256(slot0 << (256 - 24)) >> (256 - 24)),
```
kbrizzle, line 91: Yeah losing the manual formatting here is not ideal either -- I think we should have a discussion on when to use the auto-formatter. It seems like our solidity at the very least is very out of sync with what `forge fmt` generates, so I'm not sure it makes sense to make this large of a change at this point.

prateekdefi: What if we keep a combination of manual and auto formatting. We cannot skip files in forge but we can skip in prettier. @kbrizzle

prateekdefi: we can skip files like the `RiskParameter.sol` where manual formatting is required.

prateekdefi:
In team meeting decided to not go with auto formatting for now.

PR #486, opened 2024-10-25
Add 'name' and 'global' to IOracle interface

arjun-io:
The `name` and `global` getters were missing from the IOracle interface


packages/perennial-oracle/contracts/Oracle.sol
```diff
@@ -48,6 +48,12 @@ contract Oracle is IOracle, Instance {
         _updateLatest(newProvider.latest());
     }
 
+    /// @notice Returns the global state of the oracle
+    /// @return The global state of the oracle
```
EdNoepel, line 52: Better natspec would be "returns current and latest epoch of the oracle".

EdNoepel (approved):
LGTM

kbrizzle (approved):

PR #491, opened 2024-11-04
remove magic values

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1629/[cleanup]-remove-magic-values
- Removed magic values in `update()`.
- Add `close(address account, bool protect, address referrer)` method to close all position or liquidate a user
- Update `update` delta method to include `makerDelta`


packages/core/test/integration/core/liquidate.test.ts
```diff
@@ -62,7 +62,62 @@ describe('Liquidate', () => {
...
+    await chainlink.nextWithPriceModification(price => price.mul(2))
+
+    // liquidate user
+    await market.connect(userB).close(user.address, true)
+
+    expect((await market.pendingOrders(user.address, 2)).protection).to.eq(1)
+    expect(await market.liquidators(user.address, 2)).to.eq(userB.address)
+
+    // TODO: check why collateral is 999 not 1000
+    // expect((await market.locals(user.address)).collateral).to.equal(COLLATERAL)
```
prateekdefi, line 99: This expect fails in coverage but passes when this test is run with `.only`. I think this is because the deployed `market` in this test is the same as the `market` deployed in the previous test `liquidates a user`. So in a way previous test is affecting this test. Is it expected that contract states from one test are passed to the next? @EdNoepel

EdNoepel: The _Liquidate_ integration tests have a `beforeEach` which loads a fresh deployment for each test.  As such, the only statefulness might come from fakes/mocks, which are not reset along with chain state in the `loadFixture` call.  Since this is an integration test, I believe the only fakes involve the chainlink oracle setup.

prateekdefi: The problem was that chainLink params were updated in the previous test and not reset for this test, causing user collateral to reduce in liquidation. Thank you for the input. @EdNoepel

EdNoepel (commented):
Love the new `close` method and that we're moving away from the monolithic update-and-validate-invariants design.  Some comments inline.

packages/core/contracts/Market.sol
```diff
@@ -259,6 +254,36 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         _updateAndStore(context, updateContext, newOrder, newGuarantee, referrer, address(0));
     }
 
+    /// @notice Closes the account's position
+    /// @param account The account to operate on
+    /// @param protect Whether to put the account into a protected status for liquidations
+    /// @param referrer The referrer of the order
+    function close(address account, bool protect, address referrer) external {
```
kbrizzle, line 261: instead of hooking this up to the old absolute flow I think it's actually simpler to route this to the new delta update flow:
1. calculate `closable` (side not relevant yet)
2.
```
if updateContext.currentPositionLocal.maker > 0
   makerAmount = -closable
else
   if updateContext.currentPositionLocal.maker > 0
     takerAmount = -closable
   else
     takerAmount = closable
```
(can probably condense that, but just to give you an idea).
3. call `update(account, makerAmount, takerAmount, collateral, referrer)`

kbrizzle: This will have a dependency issue with https://github.com/equilibria-xyz/perennial-v2/pull/512 since the new delta update w/ maker support isn't included until there.

You might want to split these up and do them in this order instead:
1. Add `update(account, makerAmount, takerAmount, collateral, referrer)`
2. Remove magic values / add close
3. Remove absolute position

packages/periphery/test/integration/CollateralAccounts/Controller.test.ts
```diff
@@ -142,24 +143,29 @@ export function RunControllerBaseTests(
...
         expectedTo = accountA.address
-        if (amount === constants.MinInt256) expectedAmount = (await market.locals(user.address)).collateral.mul(1e12)
-        else expectedAmount = amount.mul(-1e12)
+        if (amount === constants.MinInt256) {
+          marketTransferMessage.amount = (await market.locals(user.address)).collateral.mul(-1)
+          expectedAmount = marketTransferMessage.amount.mul(-1e12)
+        } else expectedAmount = amount.mul(-1e12)
       }
 
+      const signature = await signMarketTransfer(signer, verifier, marketTransferMessage)
```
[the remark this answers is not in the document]
prateekdefi: I moved it down as `marketTransferMessage.amount` needs to be calculated before creating the signature. On looking again I think I can move it again up and remove `marketTransferMessage.amount` calculations based on Magic values in tests.

packages/core/contracts/Market.sol
prateekdefi: updated. Have moved maker delta update into this as well. This PR will implement steps 1 and 2 while #512 will implement step 3. @kbrizzle

packages/core/contracts/Market.sol
```diff
@@ -259,6 +272,45 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
+        (Context memory context, UpdateContext memory updateContext) =
+            _loadForUpdate(account, address(0), referrer, address(0), UFixed6Lib.ZERO, UFixed6Lib.ZERO);
+
+        Fixed6 makerAmount;
+        Fixed6 takerAmount;
+
+        Fixed6 closable = Fixed6Lib.from(context.latestPositionLocal.magnitude())
+            .sub(Fixed6Lib.from(context.pendingLocal.neg()));
+        
+        if (updateContext.currentPositionLocal.maker.gt(UFixed6Lib.ZERO)) {
```
kbrizzle, line 289: I kind of like this format better because you can see the result of both variables in every case:
```
        UFixed6 closable = context.latestPositionLocal.magnitude().sub(context.pendingLocal.neg());

        Fixed6 makerAmount = updateContext.currentPositionLocal.maker.isZero()
            ? Fixed6Lib.ZERO
            : Fixed6Lib.from(-1, closable);
        Fixed6 takerAmount = updateContext.currentPositionLocal.maker.isZero()
            ? updateContext.currentPositionLocal.long.isZero()
                ? Fixed6Lib.from(-1, closable)
                : Fixed6Lib.from(1, closable)
            : Fixed6Lib.ZERO;
```

that being said, the nested ternary is a little confusing, so I don't mind either format.

packages/core/contracts/types/Order.sol
```diff
@@ -83,32 +83,39 @@ library OrderLib {
...
+        bool protect,
         UFixed6 referralFee
     ) internal pure returns (Order memory newOrder) {
         newOrder.timestamp = timestamp;
         newOrder.collateral = collateral;
-        newOrder.orders = amount.isZero() ? 0 : 1;
-        newOrder.takerReferral = amount.abs().mul(referralFee);
+        newOrder.orders = makerAmount.isZero() && takerAmount.isZero() ? 0 : 1;
+        newOrder.protection = protect ? 1 : 0;
+        newOrder.takerReferral = makerAmount.abs().add(takerAmount.abs()).mul(referralFee);
```
kbrizzle, line 104: I think the `marketAmount` needs to go to the makerReferral like you originally had it?

packages/core/contracts/Market.sol
prateekdefi: I think for nested we can go with if else rather than ternary. I am keeping in if else block for this case.

packages/core/contracts/types/Order.sol
prateekdefi: It got missed while cherry picking from other PR. will update this.

kbrizzle:
Code LGTM now, couple questions on testing:
- Did we add test cases to cover the new `marketAmount` version of `update()`?
- Looks like we need to add at least the referral case for makerAmount, otherwise that would have caught the missing field.
- Is there any way to clean up the diff for `Market.test.ts`? Looks like it's having trouble finding your changes, and it's kind of hard to review.

EdNoepel (commented):
`test/unit/Order.test.ts` and `OrderTester.sol` are unchanged.  Curious why the old `Order.from` is not unit tested.  This seems easy to unit test, as we can simply pass in some position and amount scenarios and verify the _Order_ produced is as expected.

kbrizzle (approved):

prateekdefi:
> `test/unit/Order.test.ts` and `OrderTester.sol` are unchanged. Curious why the old `Order.from` is not unit tested. This seems easy to unit test, as we can simply pass in some position and amount scenarios and verify the _Order_ produced is as expected.

Added unit tests for the `Order.from`(delta)  method but not for absolute `Order.from` as it is already removed in PR #512. 
@EdNoepel

EdNoepel (approved):
LGTM

PR #492, opened 2024-11-04
Add `oracles` to IOracle and remove unused comment in `_update`

prateekdefi:
Issues addressed:
- https://linear.app/perennial/issue/PE-1917/[fix]-add-oracles-getter-to-ioracle
- https://linear.app/perennial/issue/PE-1833/[cleanup]-remove-unused-events-comments-in-market-update:


packages/perennial-oracle/contracts/interfaces/IOracle.sol
```diff
@@ -53,4 +53,5 @@ interface IOracle is IOracleProvider, IInstance {
     function withdraw(Token18 token) external;
     function beneficiary() external view returns (address);
     function global() external view returns (OracleGlobal memory);
+    function oracles(uint256) external view returns (IOracleProvider provider, uint96 timestamp);
```
EdNoepel, line 56: Returning `Epoch` struct as memory would require changes to `Oracle.sol`.  Since there are only two fields, this seems a better solution.

EdNoepel (commented):
Seems we should find a test with different oracles at different epochs and add a test confirming they can be queried through the interface.

EdNoepel (approved):
LGTM

packages/oracle/contracts/interfaces/IOracle.sol
```diff
@@ -53,4 +53,5 @@ interface IOracle is IOracleProvider, IInstance {
     function withdraw(Token18 token) external;
     function beneficiary() external view returns (address);
     function global() external view returns (OracleGlobal memory);
+    function oracles(uint256) external view returns (IOracleProvider provider, uint96 timestamp);
```
kbrizzle, line 56: It's better to surface the full struct in these kinds of getters instead of the default tuple that solidity does.

So for this one we'd want to:
- change this return type to `Epoch`
- change the storage variable to `private _oracles`
- create a getter function called `external oracles()`

packages/perennial-oracle/contracts/interfaces/IOracle.sol
kbrizzle: Oh oops looks like you guys already talked about this -- I would agree with this if we were minimizing changes in a post-audit update, however since we have the leeway to update code here, let's do it the more complete way.

kbrizzle (approved):

EdNoepel (approved):

PR #494, opened 2024-11-06
rename packages and folders

EdNoepel:
Rename packages with `v2-` prefix following Uniswap pattern.  Rename package directories to reduce redundancy; repo is is already called `perennial-v2`, so package names can be simplified to omit _perennial_ name and version.


prateekdefi (approved):
LGTM

arjun-io (approved):

kbrizzle (approved):

PR #498, opened 2024-11-15
Deprecate MultiInvoker Trigger orders

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1292/[cleanup]-deprecate-multiinvoker-trigger-orders


EdNoepel (approved):
LGTM

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
```diff
@@ -54,9 +53,6 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @dev UID for an order
     uint256 public latestNonce;
 
-    /// @dev State for the order data
-    mapping(address => mapping(IMarket => mapping(uint256 => TriggerOrderStorage))) private _orders;
```
kbrizzle, line 58: We'll need to ensure that we don't change the storage layout so that this is upgrade-compatible.

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
```diff
@@ -54,9 +53,6 @@ contract MultiInvoker is IMultiInvoker, Kept {
     /// @dev UID for an order
     uint256 public latestNonce;
```
kbrizzle, line 54: I believe these can be deprecated as well:
 - latestNonce
 - keepBufferBase
 - keepBufferCalldata

cc: @EdNoepel

EdNoepel: Good catch; I didn't realize those were only used for trigger orders.

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
```diff
@@ -21,7 +21,6 @@ import { IPythFactory } from "@perennial/v2-oracle/contracts/interfaces/IPythFac
 import { IVault } from "@perennial/v2-vault/contracts/interfaces/IVault.sol";
 import { Intent } from "@perennial/v2-core/contracts/types/Intent.sol";
 import { IMultiInvoker } from "./interfaces/IMultiInvoker.sol";
-import { TriggerOrder, TriggerOrderStorage } from "./types/TriggerOrder.sol";
```
kbrizzle, line 24: We can actually remove `Kept` from the contract as well, I think it was only used for trigger orders.

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
```diff
@@ -477,35 +385,6 @@ contract MultiInvoker is IMultiInvoker, Kept {
         return UFixed18Lib.from(raisedKeeperFee);
     }
 
-    /// @notice Places order on behalf of account from the invoker
```
kbrizzle, line 480: We can remove `_raiseKeeperFee` as this is a hook for `Kept`.

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
```diff
@@ -477,35 +385,6 @@ contract MultiInvoker is IMultiInvoker, Kept {
...
-
-    /// @notice Cancels an open order for account
-    /// @param account Account to cancel order for
-    /// @param market Market order is open in
-    /// @param nonce UID of order
-    function _cancelOrder(address account, IMarket market, uint256 nonce) internal {
-        delete _orders[account][market][nonce];
-        emit OrderCancelled(account, market, nonce);
-    }
-
```
kbrizzle, line 508: I don't we're using `marketSettle()` anymore either.

packages/periphery/contracts/MultiInvoker/interfaces/IMultiInvoker.sol
```diff
@@ -31,9 +30,6 @@ interface IMultiInvoker {
 
     event OperatorUpdated(address indexed account, address indexed operator, bool newEnabled);
     event KeeperFeeCharged(address indexed account, address indexed market, address indexed to, UFixed6 fee);
```
kbrizzle, line 32: This can be removed

packages/periphery/contracts/MultiInvoker/interfaces/IMultiInvoker.sol
```diff
@@ -44,10 +40,6 @@ interface IMultiInvoker {
     error MultiInvokerMaxFeeExceededError();
```
kbrizzle, line 40: I believe `MultiInvokerOrderMustBeSingleSidedError` and `MultiInvokerMaxFeeExceededError` can be removed.

kbrizzle (approved):

PR #500, opened 2024-11-19
Insurance Fund

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-745/[feature]-insurance-fund

- Added `InsuranceFund` contract to claim protocol fee and resolve market shortfall.


packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,68 @@
...
+import { IMarket } from "./interfaces/IMarket.sol";
+import { IMarketFactory } from "./interfaces/IMarketFactory.sol";
+import { IOracleProvider } from "./interfaces/IOracleProvider.sol";
+
+/// @title InsuranceFund
+/// @notice This contract manages the protocol fee and shortfalls for markets.
+contract InsuranceFund is IInsuranceFund, Ownable {
+
+    /// @dev The address of the market factory owner.
+    IMarketFactory public marketFactory;
```
kbrizzle, line 19: all three of these can be `immutable`, since they don't need to change / aren't set via a factory -- then they can be initialized from the constructor.

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,68 @@
...
+    function initialize(IMarketFactory _marketFactory, Token18 _token) external initializer(1) {
+        __Ownable__initialize();
+        marketFactory = _marketFactory;
+        marketFactoryOwner = marketFactory.owner();
+        DSU = _token;
+    }
+
+    /// @inheritdoc IInsuranceFund
+    function claimFees(address market) external {
+        if (market == address(0)) revert InsuranceFundInvalidAddress();
```
kbrizzle, line 37: no need for these address zero checks.

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,68 @@
...
+
+        // Check market is created from market factory
+        IOracleProvider oracle = IMarket(market).oracle();
+        if (market != address(marketFactory.markets(oracle))) revert InsuranceFundInvalidAddress();
+
+        IMarket(market).claimFee(marketFactoryOwner);
```
kbrizzle, line 43: instead of caching the market factory owner at construction, you should re-pull here each time in case ownership changes.

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,68 @@
...
+        marketFactoryOwner = marketFactory.owner();
+        DSU = _token;
+    }
+
+    /// @inheritdoc IInsuranceFund
+    function claimFees(address market) external {
+        if (market == address(0)) revert InsuranceFundInvalidAddress();
+
+        // Check market is created from market factory
+        IOracleProvider oracle = IMarket(market).oracle();
```
kbrizzle, line 40: there's an `IFactory.instances()` function you can use here instead.

packages/core/contracts/interfaces/IInsuranceFund.sol
```diff
@@ -0,0 +1,40 @@
...
+    /// @custom:error Thrown when an invalid amount is provided.
+    error InsuranceFundInvalidAmount();
+
+    /// @notice Initializes the InsuranceFund contract.
+    /// @param _marketFactory The address of the market factory.
+    /// @param _token The address of the DSU token.
+    function initialize(IMarketFactory _marketFactory, Token18 _token) external;
+
+    /// @notice Claims fees from a market.
+    /// @param market The address of the market from which to claim protocol fees.
```
kbrizzle, line 25: add a dev note here that the insurance fund must be approved as an operator of the protocol owner for claiming to work.

packages/core/contracts/interfaces/IInsuranceFund.sol
```diff
@@ -0,0 +1,40 @@
...
+    function claimFees(address market) external;
```
kbrizzle, line 26: naming -- can shorten to:
- `claim()`
- `resolve()`

packages/core/contracts/InsuranceFund.sol
kbrizzle: could also pull this check into a modifier since we need it on both claim and resolve.

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,68 @@
...
+        if (market != address(marketFactory.markets(oracle))) revert InsuranceFundInvalidAddress();
+
+        IMarket(market).claimFee(marketFactoryOwner);
+    }
+
+    /// @inheritdoc IInsuranceFund
+    function resolveShortfall(address market) external onlyOwner {
+        if (market == address(0)) revert InsuranceFundInvalidAddress();
+
+        if (IERC20(Token18.unwrap(DSU)).allowance(address(this), market) != type(uint256).max) {
```
kbrizzle, line 50: it saves gas to just always max-approve here: `DSU.approve(market)`

- either way you have to do a call out to the DSU contract, and this way you don't have to do two in the case where you need the approval.

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,68 @@
...
+
+    /// @inheritdoc IInsuranceFund
+    function sendDSUToMarket(address market, UFixed18 amount) external onlyOwner {
+        if (market == address(0)) revert InsuranceFundInvalidAddress();
+        if (amount.isZero()) revert InsuranceFundInvalidAmount();
+        DSU.push(market, amount);
+    }
+
+    /// @inheritdoc IInsuranceFund
+    function withdrawDSU(UFixed18 amount) external onlyOwner {
```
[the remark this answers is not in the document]
prateekdefi: anyone can call `claim`. Someone can claim all the fees from all markets and it might get locked in this contract if we don't have this function. What do u think?

packages/core/contracts/interfaces/IInsuranceFund.sol
kbrizzle: bump on this

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,50 @@
...
+
+/// @title InsuranceFund
+/// @notice This contract manages the protocol fee and shortfalls for markets.
+contract InsuranceFund is IInsuranceFund, Ownable {
+
+    /// @dev The address of the market factory
+    IFactory public immutable marketFactory;
+
+    /// @dev The address of DSU token
+    Token18 public immutable DSU;
```
kbrizzle, line 20: nit: maybe more generic to call this `token` since it technically could be something non-DSU.

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,50 @@
...
+        DSU = _token;
+    }
+
+    /// @inheritdoc IInsuranceFund
+    function initialize() external initializer(1) {
+        __Ownable__initialize();
+    }
+
+    /// @inheritdoc IInsuranceFund
+    function claim(address market) external isMarketInstance(IMarket(market)) {
```
kbrizzle, line 33: If you make the argument for `claim` / `resolve`: `IMarket market` I think you'll have to do less casting throughout the code.

packages/core/contracts/InsuranceFund.sol
```diff
@@ -0,0 +1,50 @@
...
+    /// @inheritdoc IInsuranceFund
+    function resolve(address market) external onlyOwner isMarketInstance(IMarket(market)) {
+        DSU.approve(market);
+        IMarket(market).claimExposure();
+    }
+
+    /// @notice Validates that a market was created by the market factory
+    /// @param market Market to validate
+    modifier isMarketInstance(IMarket market) {
+        // Check market is created from market factory
```
kbrizzle, line 46: the natspec probably covers this comment

packages/core/contracts/interfaces/IMarket.sol
```diff
@@ -171,4 +171,5 @@ interface IMarket is IInstance {
     function updateParameter(MarketParameter memory newParameter) external;
     function updateRiskParameter(RiskParameter memory newRiskParameter) external;
     function claimFee(address account) external returns (UFixed6);
+    function claimExposure() external;
```
kbrizzle, line 174: let's not add this in this PR (we're also removing this functionality anyways in a separate v2.4 task)

packages/core/contracts/interfaces/IInsuranceFund.sol
```diff
@@ -0,0 +1,27 @@
...
+    function initialize() external;
+
+    /// @notice Claims fees from a market.
+    /// @param market The address of the market from which to claim protocol fees.
+    function claim(address market) external;
+
+    /// @notice Resolves any shortfall in a market.
+    /// @param market The address of the market for which to resolve the shortfall.
+    function resolve(address market) external;
+}
```
kbrizzle, line 27: nit: newline at end

kbrizzle:
Could you add integration tests for this as well? I know that's much harder in this case, but it would be good to test the full claim / resolve flow to ensure all of the permissioning actually works when everything is setup.

packages/core/contracts/interfaces/IMarket.sol
prateekdefi: Added this to `IMarket` to call Market's `claimExposure` from the `InsuranceFund` contract.

packages/core/test/integration/core/happyPath.test.ts
```diff
@@ -1479,6 +1480,92 @@ describe('Happy Path', () => {
...
+      maxPriceDeviation: parse6decimal('0.1'),
+      closed: false,
+      settle: false,
+    }
+
+    await market.updateParameter(parameter)
+    await market.updateRiskParameter(riskParameter)
+
+    // ensure exposure is negative
+    expect((await market.global()).exposure).to.lt(0)
```
EdNoepel, line 1558: nit: I try to fashion the chai matchers to read like English sentences, for example `to.be.lt(0)` and on line 1566 `to.equal(0)`.  Entirely cosmetic though.

EdNoepel (approved):
LGTM

packages/core/contracts/interfaces/IMarket.sol
prateekdefi: Removed in [c358fd3](https://github.com/equilibria-xyz/perennial-v2/pull/500/commits/c358fd3a0b446e7931537f722c20f005bf9d0d18), since we no longer call `claimExposure`.

packages/core/test/integration/core/fees.test.ts
```diff
@@ -2388,6 +2388,55 @@ describe('Fees', () => {
...
+        oracleFee: expectedOracleFee,
+        riskFee: expectedRiskFee,
+        latestPrice: parse6decimal('113.882975'),
+      })
+
+      // set insurance fund as operator for market factory owner
+      await marketFactory.connect(owner).updateOperator(insuranceFund.address, true)
+
+      // claim protocol fee
+      await insuranceFund.connect(owner).claim(market.address)
```
EdNoepel, line 2438: Shouldn't we expect this to not revert, check emitted events, check balance of insurance fund?

packages/core/test/integration/core/liquidate.test.ts
```diff
@@ -135,6 +136,75 @@ describe('Liquidate', () => {
...
+    expect((await market.locals(user.address)).collateral).to.equal(BigNumber.from('-2524654460'))
+
+    await marketFactory.connect(owner).updateExtension(insuranceFund.address, true)
+
+    const insuranceFundSigner = await impersonate(insuranceFund.address)
+
+    await fundWallet(dsu, insuranceFundSigner)
+
+    // resolve the shortfall
+    await insuranceFund.connect(owner).resolve(market.address, user.address)
```
EdNoepel, line 204: Recommend either checking _Market_ emits the an expected `OrderCreated` event here or, since locals are checked below, maybe just that it didn't revert.

EdNoepel (approved):
LGTM

kbrizzle (approved):

PR #504, opened 2024-11-26
Update `staleAfter` to uint16

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1978/[cleanup]-make-staleafter-uint16

### DIFF
--- a/packages/core/contracts/types/RiskParameter.sol
+++ b/packages/core/contracts/types/RiskParameter.sol
@@ -84,7 +84,7 @@ using RiskParameterStorageLib for RiskParameterStorage global;
 ///        uint48 minMargin;                           // <= 281m
 ///        uint48 minMaintenance;                      // <= 281m
 ///        uint32 liquidationFee;                      // <= 4294
-///        uint24 staleAfter;                          // <= 16m s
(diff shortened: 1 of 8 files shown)


kbrizzle (approved):

packages/core/contracts/types/RiskParameter.sol
```diff
@@ -84,7 +84,7 @@ using RiskParameterStorageLib for RiskParameterStorage global;
 ///        uint48 minMargin;                           // <= 281m
 ///        uint48 minMaintenance;                      // <= 281m
 ///        uint32 liquidationFee;                      // <= 4294
-///        uint24 staleAfter;                          // <= 16m s
+///        uint16 staleAfter;                          // <= 65536 seconds
```
EdNoepel, line 87: I would have commented _18.2 hours_, but an improvement nonetheless.

EdNoepel (approved):
LGTM

packages/core/contracts/types/RiskParameter.sol
prateekdefi: I have used seconds for consistency, earlier it was written 16million seconds.

PR #507, opened 2024-12-02
Manage collateral through Margin contract

EdNoepel:
## Background
To implement cross-margin, all collateral will be managed by a new `Margin` contract.  Markets will no longer hold any token balance.  As a step towards this goal, this PR introduces the new contract and uses it to manage collateral in "isolated" mode.  Calling `IMarket.update` and passing nonzero values for the collateral argument no longer moves tokens.  Instead, a positive value allocates more collateral to isolated mode (putting the market into isolated mode if appropriate).  A negative value allocates less collateral from isolated mode, which will (eventually) make it available for cross-margin.

## Purpose and Scope
Goal of this PR is to get _Margin_ contract in place in the `v2.4` branch such that cross-margin facilities may be implemented.  Until then, tests will use isolated collateral only.

Explicitly out-of-scope for this PR:
- Actually implementing cross-margin functionality
- Interactions through signed messages
- Rolling the _Margin_ contract into _MarketFactory_

## TODO
- [x] Cleanup commits: remove commented-out console statements and such
- [x] Test `multi-delayed update w/ collateral (gas)`
- [x] Check coverage - Only missing coverage is for (unimplemented) cross-margin functionality.
- [x] Address (possibly unrelated) linter warnings


packages/core/contracts/Market.sol
```diff
@@ -332,7 +339,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         _locals[account].store(newLocal);
 
         if (!feeReceived.isZero()) {
-            token.push(msg.sender, UFixed18Lib.from(feeReceived));
+            margin.updateBalance(msg.sender, Fixed6Lib.from(feeReceived));
```
kbrizzle, line 342: this is good for now, but we'll probably want to push all these fee balances and claim flows into `Margin` as well, instead of still triggering stuff from the Market.

packages/core/contracts/Market.sol
```diff
@@ -342,9 +349,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     function claimExposure() external onlyOwner {
         Global memory newGlobal = _global.read();
 
-        if (newGlobal.exposure.sign() == 1) token.push(msg.sender, UFixed18Lib.from(newGlobal.exposure.abs()));
-        if (newGlobal.exposure.sign() == -1) token.pull(msg.sender, UFixed18Lib.from(newGlobal.exposure.abs()));
-
+        margin.updateBalance(msg.sender, newGlobal.exposure);
```
kbrizzle, line 352: don't worry about this one too much, we'll be removing `claimExposure()` and `exposure` more generally in v2.4 with the price impact work.

packages/core/contracts/Market.sol
```diff
@@ -426,11 +431,11 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
     }
 
-    /// @notice Returns the local checkpoint for the given account and version
-    /// @param account The account to query
+    /// @notice Returns the local isolated checkpoint for the given account and version
+    /// @param account The account with isolated collateral to query
     /// @param version The version to query
     function checkpoints(address account, uint256 version) external view returns (Checkpoint memory) {
-        return _checkpoints[account][version].read();
+        return margin.isolatedCheckpoints(account, this, version);
```
kbrizzle, line 438: fine to leave this for now, but we may way to redirect stuff to point at the native Margin getters eventually.

packages/core/contracts/Market.sol
```diff
@@ -665,8 +670,8 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         updateContext.guaranteeGlobal.add(newGuarantee);
         updateContext.guaranteeLocal.add(newGuarantee);
 
-        // update collateral
-        context.local.update(newOrder.collateral);
+        // update collateral in margin contract
+        margin.handleMarketUpdate(context.account, newOrder.collateral);
```
kbrizzle, line 674: per lower comments in `InvariantLib`

We may want to split this up so that the overall `update()` flow is:
1. `Margin.isolate()`
2. `Market._update()`
3. `Margin.checkMargin()`

Up to you on whether that's too much of a change for the first pass since this does look more "in-place"

packages/core/contracts/Market.sol
```diff
@@ -856,10 +861,10 @@ contract Market is IMarket, Instance, ReentrancyGuard {
             versionTo
         );
 
-        context.local.update(newOrderId, accumulationResponse);
         context.latestPositionLocal.update(newOrder);
-
-        _checkpoints[context.account][newOrder.timestamp].store(settlementContext.latestCheckpoint);
+        // calculate and store collateral change for account
+        Fixed6 pnl = context.local.update(newOrderId, accumulationResponse);
+        margin.updateCheckpoint(context.account, newOrder.timestamp, settlementContext.latestCheckpoint, pnl);
```
kbrizzle, line 867: we'll want to eventually pass in `claimable` stuff as well here (global fees, liquidation fee, referral), but feel to leave this in Market for now if that's easier.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+    /// @dev Contract used to validate markets
+    IMarketFactory public marketFactory;
+
+    /// @notice Collateral spread across markets: user -> balance
+    mapping(address => Fixed6) public crossMarginBalances;
+
+    // TODO: Introduce an iterable collection of cross-margained markets for a user.
+
+    /// @notice Non-cross-margained collateral: user -> market -> balance
+    mapping(address => mapping(IMarket => Fixed6)) public isolatedBalances;
```
kbrizzle, line 30: you can use `address(0)` for `IMarket` as the cross margin balance.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+
+    /// @dev Storage for isolated account checkpoints: user -> market -> version -> checkpoint
+    mapping(address => mapping(IMarket => mapping(uint256 => CheckpointStorage))) private _isolatedCheckpoints;
```
kbrizzle, line 33: same thing here for future reference, the cross-margin checkpoint can be stored at `IMarket` -> `address(0)`.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+        DSU.pull(msg.sender, UFixed18Lib.from(amount));
+        crossMarginBalances[account] = crossMarginBalances[account].add(Fixed6Lib.from(amount));
+        emit FundsDeposited(account, amount);
+    }
+
+    // TODO: support a magic number for full withdrawal?
+    /// @inheritdoc IMargin
+    function withdraw(address account, UFixed6 amount) external nonReentrant onlyOperator(account) {
+        Fixed6 balance = crossMarginBalances[account];
+        if (balance.lt(Fixed6Lib.from(amount))) revert MarginInsufficientCrossedBalance();
```
kbrizzle, line 62: note: if we do the shortfall solution where we keep a separate per-market shortfall tally instead of letting the user's collateral go negative, then we can clean this up.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+        DSU.push(msg.sender, UFixed18Lib.from(amount));
+        emit FundsWithdrawn(account, amount);
+    }
+
+    /// @inheritdoc IMargin
+    function isolate(address account, IMarket market) external nonReentrant onlyOperator(account){
+        if (!_isIsolated(account, market)) revert MarginMarketNotCrossed();
+
+        market.settle(account);
+        if (_hasPosition(account, market)) revert MarginHasPosition();
```
kbrizzle, line 74: wrote some notes in `InvariantLib` about this, but I think we need a couple things here:

- A "switching modes" check, similar to the single-sided check in Market, but for isolated <> cross
  - This is a little more advanced than just checking if there's a position open
- We'll need to verify the margin as well (for withdrawals)

Would be ideal to kind of figure out a pattern here where we get this all for free, since this function would also be called as part of the Market.update() flow.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+        // TODO: update collections which track which markets are isolated/crossed
+        emit MarketIsolated(account, market);
+    }
+
+    /// @inheritdoc IMargin
+    function adjustIsolatedBalance(
+        address account,
+        IMarket market,
+        Fixed6 amount
+    ) external nonReentrant onlyOperator(account) {
```
kbrizzle, line 85: Per notes on the linear ticket, I think we can combine this function with `isolate()`.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+        checkpoint.collateral = checkpoint.collateral.add(amount);
+        // console.log("adjustIsolatedBalance storing checkpoint with collateral %s at %s",
+        //     UFixed6.unwrap(checkpoint.collateral.abs()), latestTimestamp);
+        _isolatedCheckpoints[account][market][latestTimestamp].store(checkpoint);
+
+        emit IsolatedFundsChanged(account, market, amount);
+    }
+
+    /// @inheritdoc IMargin
+    function cross(address account, IMarket market) external nonReentrant onlyOperator(account) {
```
kbrizzle, line 113: Per notes on the linear ticket, I think we remove this.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /// @inheritdoc IMargin
+    function maintained(address account) external returns (bool isMaintained) {
+        // TODO: settle all markets and check maintenance requirements
+        isMaintained = false;
+    }
+
+    /// @inheritdoc IMargin
+    function margined(address account) external returns (bool isMargined) {
```
kbrizzle, line 135: I think these two are duplicates of `checkMargined` and `checkMaintained`? I like these namings better though.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+            //       when aggregating sender market, use positionMagnitude and latestVersion provided by caller
+            revert("Cross-margin not yet implemented");
+        }
+    }
+
+
+
+    // TODO: we won't need this if we remove collateral params from Market update methods.
+    /// @inheritdoc IMargin
+    function handleMarketUpdate(address account, Fixed6 collateralDelta) external onlyMarket {
```
kbrizzle, line 180: In `Market` you're using this for the isolation amount - if that's correct, we should try and combine this with `isolate` so that we can call that instead of having a custom flow.

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+        if (newIsolatedBalance.lt(Fixed6Lib.ZERO)) revert MarginInsufficientIsolatedBalance();
+
+        crossMarginBalances[account] = newCrossBalance;
+        isolatedBalances[account][market] = newIsolatedBalance;
+
+        // TODO: Ensure InvariantLib checks margin and maintenance requirements and reverts where appropriate.
+    }
+
+    /// @inheritdoc IMargin
+    function updateBalance(address account, Fixed6 collateralDelta) external onlyMarket {
```
kbrizzle, line 215: I think we only need one between `updateBalance` and `updateCheckpoint `

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,316 @@
...
+        _updateCollateralBalance(account, IMarket(msg.sender), collateralDelta);
+        // TODO: Emit an event
+    }
+
+    /// @inheritdoc IMargin
+    function updateCheckpoint(address account, uint256 version, Checkpoint memory latest, Fixed6 pnl) external onlyMarket{
+        // console.log("updateCheckpoint storing checkpoint with collateral %s at %s",
+        //     UFixed6.unwrap(latest.collateral.abs()), version);
+        // Store the checkpoint
+        _isolatedCheckpoints[account][IMarket(msg.sender)][version].store(latest);
```
kbrizzle, line 225: We'll want to move in the actually updating of the `Checkpoint` object into Margin eventually as well, instead of updating it in `Market` and passing the new one in. (seeing as the `pnl` data is duplicative at the moment).

packages/core/contracts/Margin.sol
EdNoepel: Done in 9e2dac2ff7924211c67587d034aafd3e74c006bd

packages/core/contracts/Margin.sol
EdNoepel: Done in 9e2dac2ff7924211c67587d034aafd3e74c006bd

packages/core/contracts/Market.sol
```diff
@@ -19,26 +20,30 @@ import { Local, LocalStorage } from "./types/Local.sol";
...
 
+    // TODO: Any reason to retain this, or should we remove from MarketDefinition?
     /// @dev The underlying token that the market settles in
     Token18 public token;
 
-    /// @dev DEPRECATED SLOT -- previously the reward token
-    bytes32 private __unused0__;
+    // TODO: Is okay that this storage slot is reused, and that it is preceeded by a non-immutable field?
+    /// @dev Handles collateral across all markets
+    IMargin public immutable margin;
```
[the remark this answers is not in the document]
EdNoepel: Indeed; cleaned up in 9e2dac2ff7924211c67587d034aafd3e74c006bd

packages/core/contracts/types/Local.sol
```diff
@@ -83,14 +76,13 @@ library LocalStorageLib {
...
-        if (newValue.collateral.gt(Fixed6.wrap(type(int64).max))) revert LocalStorageInvalidError();
-        if (newValue.collateral.lt(Fixed6.wrap(type(int64).min))) revert LocalStorageInvalidError();
+        if (!newValue.collateral.eq(Fixed6Lib.ZERO)) revert LocalStorageInvalidError();
         if (newValue.claimable.gt(UFixed6.wrap(type(uint64).max))) revert LocalStorageInvalidError();
 
         uint256 encoded0 =
             uint256(newValue.currentId << (256 - 32)) >> (256 - 32) |
             uint256(newValue.latestId << (256 - 32)) >> (256 - 32 - 32) |
-            uint256(Fixed6.unwrap(newValue.collateral) << (256 - 64)) >> (256 - 32 - 32 - 64) |
+            uint256(0 << (256 - 64)) >> (256 - 32 - 32 - 64) |
```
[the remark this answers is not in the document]
EdNoepel: Done in 9e2dac2ff7924211c67587d034aafd3e74c006bd

packages/core/contracts/Margin.sol
EdNoepel: Done in 60bac95542f4987d47600079bb84997fe4c08d79

packages/core/contracts/Margin.sol
EdNoepel: Done in 60bac95542f4987d47600079bb84997fe4c08d79

packages/core/contracts/Margin.sol
EdNoepel: _Market.claimFee_ and _Market.claimExposure_ do not currently publish a checkpoint.  With the new implementation, should we store a new checkpoint with latest version and adjusted collateral balance, even though no settlement occurred?

kbrizzle: The `claimable` parallel balance was originally created to track these "out of settlement / checkpoint flow" balances. Let's leave that in for now (probably move this state to the Margin contract tho?). That way we don't need to make major updates to the places we're taking checkpoints.

Exposure works differently, but we're removing that one, so we don't need to worry about the implications here.

packages/core/contracts/Margin.sol
EdNoepel: Indeed; done in 958d459f87347f41d348b15c9f2151bb69051cb2.

EdNoepel: Some complexities:
- We would need _isolate_ to allow `msg.sender` to be the _Market_.  We could handle this with a new modifier just for this method.
- _Margin.isolate_ performs a settlement on the _Market_, which would trigger a reentrancy revert.  But we don't want to implicitly settle when a user calls _update_.
- Currently, no checkpoint is written when collateral is updated for an empty order.  Unsure what the expected behavior is, because direct calls to _isolate_ will currently produce a checkpoint, but tests which use _Market.update_ expect no checkpoint until settlement.

I refactored the implementation in 0b837480e20ef7ea789f0b8fcbeff50d0338ba86 to clean this up a bit.

packages/core/contracts/Margin.sol
EdNoepel: Per the design doc, `margined` and `maintained` methods settle all markets (crossed and isolated) and check everything.  It would be gas inefficient to perform unnecessary iteration when a market in isolated collateral mode merely needs to check whether margin/maintenance requirements are met for a user.

The _check*_ methods were created to provide something more appropriate for _InvariantLib_ or _Market_ to call.

packages/core/contracts/libs/InvariantLib.sol
```diff
@@ -77,7 +80,7 @@ library InvariantLib {
         if (
             !updateContext.signer &&                                            // sender is relaying the account's signed intention
             !updateContext.operator &&                                          // sender is operator approved for account
-            !(newOrder.isEmpty() && newOrder.collateral.gte(Fixed6Lib.ZERO))    // sender is depositing zero or more into account, without position change
+            !(newOrder.isEmpty() && newOrder.collateral.gte(Fixed6Lib.ZERO))    // sender is isolating collateral into account, without position change
```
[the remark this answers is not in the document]
EdNoepel: When _Market_ calls into _Margin_ contract, we lose the `msg.sender`, which becomes the _Market_ contract.  We could remove this by adding a _sender_ argument to `handleMarketUpdate` and duplicating the authorization logic, but leaving this here seems cleaner.

packages/core/contracts/libs/InvariantLib.sol
```diff
@@ -37,17 +39,18 @@ library InvariantLib {
 
         if (newOrder.protected() && (
             !context.pendingLocal.neg().eq(context.latestPositionLocal.magnitude()) ||  // total pending close is not equal to latest position
-            context.latestPositionLocal.maintained(                                     // latest position is properly maintained
-                context.latestOracleVersion,
-                context.riskParameter,
-                context.local.collateral
+            margin.checkMaintained(                                                     // latest position is properly maintained
```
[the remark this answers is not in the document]
EdNoepel: I did this, such that the check occurs in _Market_ which makes a call to _Margin_ which queries requirements back from _Market_ again.  This is less gas-efficient, as we have to re-read the position (1 slot) and pending order (3 slots) from storage, rather than reading them from the context.  Thankfully this only happens once because margin is not checked in protected mode, and maintenance is not checked outside of protected mode.

packages/core/contracts/libs/InvariantLib.sol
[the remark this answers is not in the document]
EdNoepel: This has been refactored a bit; there is now an operator check in _Market_ for collateral movement, and the one in _InvariantLib_ just cares about non-empty orders.

packages/core/contracts/libs/InvariantLib.sol
```diff
@@ -37,17 +39,18 @@ library InvariantLib {
...
             ) ||
-            !newOrder.collateral.eq(Fixed6Lib.ZERO)                                     // the order is modifying collateral
+            !newOrder.collateral.eq(Fixed6Lib.ZERO)                                     // the order is modifying isolated collateral
         )) revert IMarket.MarketInvalidProtectionError();
 
+        // TODO: This needs to move to Margin contract
         if (
             !(updateContext.currentPositionLocal.magnitude().isZero() && context.latestPositionLocal.magnitude().isZero()) &&       // sender has no position
-            !(newOrder.isEmpty() && newOrder.collateral.gte(Fixed6Lib.ZERO)) &&                                                     // sender is depositing zero or more into account, without position change
+            !(newOrder.isEmpty() && newOrder.collateral.gte(Fixed6Lib.ZERO)) &&                                                     // sender is isolating collateral into account, without position change
```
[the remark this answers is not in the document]
EdNoepel: Rather than moving `MarketStalePriceError`, it seems we should split it into to:
	- _InvariantLib_ continues to revert if price is stale when user has a position and is trying to change their position (non-empty order).
	- _Margin_ reverts if price is stale when user has a position and is trying to de-isolate collateral from the market.
@kbrizzle  Could you confirm this logic?

packages/core/contracts/interfaces/IMarket.sol
```diff
@@ -21,8 +22,9 @@ import { VersionAccumulationResult } from "../libs/VersionLib.sol";
 import { CheckpointAccumulationResult } from "../libs/CheckpointLib.sol";
 
 interface IMarket is IInstance {
+    // TODO: Shall we eliminate this in favor of just passing IOracleProvider?
+    // Or are more market initialization args expected?
```
EdNoepel, line 26: @kbrizzle Your thoughts on this would be appreciated.

kbrizzle: Yeah good idea, let's get rid of this -- this originally had ~5 fields, so now that we're at one, it's no longer necessary.

packages/core/contracts/libs/InvariantLib.sol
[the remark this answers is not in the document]
kbrizzle: Hmm yeah I see what you're saying, but it would be nice for there to be a single source of truth for staleness.

My logic was that staleness prevents a valid margin / maintenance reading which would fail the margin invariant for both of these cases.

I think the ideal setup would be to combine the staleness check w/ the margin and maintenance checks somehow, but I'm not really sure offhand the cleanest way to hooking all that up. For now if this makes more sense feel free to go with it, these cleanup refactors will be easier to reason about once we've push the first chunk of changes.

packages/core/contracts/Margin.sol
kbrizzle: Got it ok, I think this is good enough for now, similar to my other comments we can take another look at this later on to see if there's a cleaner way of hooking it up.

packages/core/contracts/Market.sol
```diff
@@ -426,11 +429,31 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
-        return _checkpoints[account][version].read();
+        return margin.isolatedCheckpoints(account, this, version);
+    }
+
+    /// @inheritdoc IMarket
+    function maintenanceRequired(
+        address account
+    ) external view returns (UFixed6 requirement) {
+        (OracleVersion memory latestOracleVersion, ) = oracle.status();
+        UFixed6 positionMagnitude = _positions[account].read().magnitude();
```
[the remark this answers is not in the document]
EdNoepel: Doing so breaks the `reverts if position increases in magnitude` test, as userB is no longer liquidable if we count their pending long order.  Would you like me to write up a separate ticket for this?

packages/core/contracts/interfaces/IMarket.sol
EdNoepel: Done in e39fa4bf5f018b8874226d7a38138694c53aa62c.

packages/core/contracts/libs/InvariantLib.sol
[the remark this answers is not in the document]
EdNoepel: Yesterday's revision moves logic to a new `Market._stale` method exposed through `IMarket`, called by `Margin` contract where appropriate.  I put the three checks adjacent to each other, after state is saved, to facilitate further refactoring.

packages/core/contracts/Market.sol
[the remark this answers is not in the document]
EdNoepel: Created PE-2067 to track this.

packages/core/contracts/libs/InvariantLib.sol
```diff
@@ -37,17 +39,18 @@ library InvariantLib {
...
-                context.latestOracleVersion,
-                context.riskParameter,
-                context.local.collateral
+            margin.checkMaintained(                                                     // latest position is properly maintained
+                context.account,
+                context.latestPositionLocal.magnitude(),
+                context.latestOracleVersion
             ) ||
-            !newOrder.collateral.eq(Fixed6Lib.ZERO)                                     // the order is modifying collateral
+            !newOrder.collateral.eq(Fixed6Lib.ZERO)                                     // the order is modifying isolated collateral
```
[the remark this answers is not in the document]
EdNoepel: The current implementation of `InvariantLib` does not look at collateral anywhere.  Delayed orders are outside the scope of this PR; all changes to isolated collateral (positive and negative) are processed prior to the update.  Maintenance/margin requirements are checked after the update.

packages/core/contracts/libs/InvariantLib.sol
```diff
@@ -86,12 +89,12 @@ library InvariantLib {
...
+            !margin.checkMargained(
+                context.account,
                 context.latestPositionLocal.magnitude().add(context.pendingLocal.pos()),
                 context.latestOracleVersion,
-                context.riskParameter,
                 updateContext.collateralization,
-                context.local.collateral.add(newGuarantee.priceAdjustment(context.latestOracleVersion.price)) // apply price override adjustment from intent if present
+                newGuarantee.priceAdjustment(context.latestOracleVersion.price) // apply price override adjustment from intent if present
             )
         ) revert IMarket.MarketInsufficientMarginError();
```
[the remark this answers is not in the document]
EdNoepel: `InvariantLib` does not look at collateral any more.  `Market._updateAndStore` currently performs the maintenance check after processing the update.

packages/core/contracts/Market.sol
EdNoepel: If we drove this from `Margin` contract, `Market` could restrict this method as `onlyMargin` and then _Margin_ could pull the DSU afterward.  Was that what you envisioned?

packages/core/contracts/Market.sol
EdNoepel: The current flow is:
1. `Margin._isolate()` (through `margin.handleMarketUpdate`)
2. `Market._update()`
3. `Market._storeContext()`
4. `Margin.maintained()` or `Margin.margined()`

packages/core/contracts/Margin.sol
```diff
@@ -0,0 +1,248 @@
...
+        _balances[account][CROSS_MARGIN] = newCrossBalance;
+        _balances[account][market] = newIsolatedBalance;
+        if (updateCheckpoint_) {
+            uint256 latestTimestamp = market.oracle().latest().timestamp;
+            Checkpoint memory checkpoint = _checkpoints[account][market][latestTimestamp].read();
+            checkpoint.collateral = checkpoint.collateral.add(amount);
+            _checkpoints[account][market][latestTimestamp].store(checkpoint);
+        }
+
+        // TODO: not sure how I feel about Margin contract reverting with Market errors here
```
EdNoepel, line 193: Seems better for DX to reuse the error rather than create a new one consumers need to worry about.  Removing this comment.

packages/core/contracts/types/Local.sol
```diff
@@ -83,14 +76,12 @@ library LocalStorageLib {
     function store(LocalStorage storage self, Local memory newValue) internal {
         if (newValue.currentId > uint256(type(uint32).max)) revert LocalStorageInvalidError();
         if (newValue.latestId > uint256(type(uint32).max)) revert LocalStorageInvalidError();
-        if (newValue.collateral.gt(Fixed6.wrap(type(int64).max))) revert LocalStorageInvalidError();
-        if (newValue.collateral.lt(Fixed6.wrap(type(int64).min))) revert LocalStorageInvalidError();
+        if (!newValue.collateral.eq(Fixed6Lib.ZERO)) revert LocalStorageInvalidError();
```
EdNoepel, line 79: Seems valueable to trap anything trying to store a nonzero value rather than deleting the line.

packages/core/test/integration/core/fees.test.ts
```diff
@@ -2629,12 +2689,10 @@ describe('Fees', () => {
...
         latestId: 1,
-        collateral: COLLATERAL.sub(TRADE_FEE_D)
-          .sub(TAKER_LINEAR_FEE)
-          .sub(TAKER_PROPORITIONAL_FEE)
-          .sub(TAKER_ADIABATIC_FEE)
-          .sub(14), // loss of precision
       })
+      expect(await margin.isolatedBalances(userD.address, market.address)).to.equal(
+        COLLATERAL.sub(TRADE_FEE_D).sub(TAKER_LINEAR_FEE).sub(TAKER_PROPORITIONAL_FEE).sub(TAKER_ADIABATIC_FEE).sub(14),
+      ) // loss of precision
```
EdNoepel, line 2695: I had separated each operation such that the `// loss of precision` comment was adjacent to the `.sub(14)`.  However, with the outdent, the auto-linter saw an opportunity.

packages/core/test/unit/market/Market.test.ts
```diff
@@ -388,13 +384,6 @@ async function settle(market: Market, account: SignerWithAddress, sender?: Signe
   return await market.connect(sender || account).settle(account.address)
 }
 
-async function deposit(market: Market, amount: BigNumber, account: SignerWithAddress, sender?: SignerWithAddress) {
-  const accountPositions = await market.positions(account.address)
-  return await market
-    .connect(sender || account)
-    ['update(address,int256,int256,address)'](account.address, 0, amount, constants.AddressZero)
-}
-
```
EdNoepel, line 397: @prateekdefi  and I both created a `deposit` function in separate branches.  Coalesced logic for both on line 413 such that both contracts do not need to be passed in as params.

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
```diff
@@ -197,8 +201,10 @@ contract MultiInvoker is IMultiInvoker, Initializable {
             interfaceFee1.receiver == address(0) ? interfaceFee2.receiver : interfaceFee1.receiver
         );
 
-        Fixed6 withdrawAmount = Fixed6Lib.from(Fixed18Lib.from(DSU.balanceOf()).sub(balanceBefore));
```
EdNoepel, line 200: This way of calculating the amount withdrawn will no longer work, because the `market.update()` call on line 194 does not result in a transfer of funds.

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
```diff
@@ -246,20 +252,19 @@ contract MultiInvoker is IMultiInvoker, Initializable {
...
     /// @param target Market or Vault to approve
     function _approve(address target) internal {
-        if (
-            !marketFactory.instances(IInstance(target)) &&
-            !vaultFactory.instances(IInstance(target))
-        ) revert MultiInvokerInvalidInstanceError();
-
-        DSU.approve(target);
+        if (marketFactory.instances(IInstance(target))) {
+            DSU.approve(address(IMarket(target).margin()));
```
EdNoepel, line 262: Toiled over this a bit.  We don't have a way to verify the authenticity of a _Margin_ contract, but we can verify a _Market_ is legitimate.  By continuing to have the consumer approve their Markets, we ensure the correct _Margin_ contract gets approved, and avoid changing the consumer's approval process.

packages/vault/contracts/Vault.sol
```diff
@@ -167,9 +168,12 @@ contract Vault is IVault, Instance {
     /// @param market The market to register
     function _register(IMarket market) private {
         if (!IVaultFactory(address(factory())).marketFactory().instances(market)) revert VaultNotMarketError();
-        if (!market.token().eq(asset)) revert VaultIncorrectAssetError();
+        // TODO: Since MarketFactory is the same, is the second check superfluous?
+        // This vault could have been created with an old Market implementation which uses the same token but
+        // a different Margin contract.  Maybe create a new error (and test) for this case.
+        if (!market.margin().DSU().eq(asset) || market.margin() != margin) revert VaultIncorrectAssetError();
```
EdNoepel, line 174: I don't think this test requires a new error, but would be nice to have a test for it.

packages/core/contracts/Market.sol
```diff
@@ -622,11 +657,44 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
         // store updated state
         _storeContext(context);
+
+        // TODO: consider updating these three checks into a post-save invariant;
+        // perhaps moving into a new InvariantLib function to reduce contract size
+
+        // when liquidating, ensure maintenance requirements are not met
+        if (newOrder.protected() && (
+            !context.pendingLocal.neg().eq(context.latestPositionLocal.magnitude()) || // total pending close is not equal to latest position
+            margin.maintained(context.account)                                    // latest position is properly maintained
```
kbrizzle, line 681: we need to verify that no collateral change is occurring for a valid protection: https://github.com/equilibria-xyz/perennial-v2/pull/507/files#diff-ba3bd1ab3098202f47f9a14185da121b85273a71d061a108f2f482c4ae6aa19bL45.

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
kbrizzle: I think we may be able to deprecate this flow now and basically have to separate UX steps:

1. Deposit / Withdraw + Wrap / Unwrap to cross margin account
2. Update via cross margin account or deposit + update via isolated account

cc: @arjun-io

packages/periphery/contracts/MultiInvoker/MultiInvoker.sol
kbrizzle: Actually now that we just have one collateral entry point we could approve the `Margin` contract in this setup via a parameterized address, and then deprecate this entirely.

packages/vault/contracts/Vault.sol
```diff
@@ -501,6 +509,8 @@ contract Vault is IVault, Instance {
 
             // local
             Local memory local = registration.market.locals(address(this));
+            // TODO: maybe shouldn't be reusing a field we're trying to deprecate
```
kbrizzle, line 512: agree with this 😂

packages/core/contracts/Margin.sol
kbrizzle: To clarify here it is not safe to pass claimable into `collateralDelta` here because it will not properly checkpoint the newly deposit collateral (I think).

We should keep a separate `claimable` balance under the user's account like we did in Market and update that instead.

kbrizzle:
Please take a look at: https://github.com/equilibria-xyz/perennial-v2/pull/507#discussion_r1902280945 and https://github.com/equilibria-xyz/perennial-v2/pull/507#discussion_r1902287242, but all of the others can just be noted for your follow up refactor(s).

packages/core/contracts/Market.sol
```diff
@@ -936,6 +1001,12 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         }
     }
 
+    /// @dev Returns true if the oracle price is stale, which should prevent position change and deisolation of collateral
+    function _stale() private view returns (bool isStale) {
+        (OracleVersion memory latest, uint256 currentTimestamp) = oracle.status();
+        isStale = !latest.valid || currentTimestamp - latest.timestamp >= _riskParameter.read().staleAfter;
```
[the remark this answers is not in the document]
EdNoepel: changed in 0da5b45e2c4f0ef40144efb11fcb0c722ee0fc70

packages/core/contracts/Market.sol
EdNoepel: Good catch.  This only applies to self-liquidation since non-operators are never allowed to change isolated collateral.  Fixed and added test in 0da5b45e2c4f0ef40144efb11fcb0c722ee0fc70.

packages/core/contracts/Market.sol
```diff
@@ -936,6 +1001,12 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         }
     }
 
+    /// @dev Returns true if the oracle price is stale, which should prevent position change and deisolation of collateral
+    function _stale() private view returns (bool isStale) {
```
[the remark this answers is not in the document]
EdNoepel: Thought we wanted to move away from using context as we did when checking maintenance/margin requirements.  I prefer reusing values in the context; parameterized in 4e6948d5e50b73d918e261ee03e7928eb6e7a03c.

packages/core/contracts/Market.sol
```diff
@@ -477,11 +479,44 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
     function checkpoints(address account, uint256 version) external view returns (Checkpoint memory) {
-        return _checkpoints[account][version].read();
+        return margin.isolatedCheckpoints(account, this, version);
+    }
+
+    /// @inheritdoc IMarket
+    function hasPosition(address account) external view returns (bool) {
+        Position memory position_ = _positions[account].read();
+        Order memory pending_ = _pendings[account].read();
+        return !(position_.magnitude().isZero() && pending_.isEmpty());
```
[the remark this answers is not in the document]
EdNoepel: Done in 4e6948d5e50b73d918e261ee03e7928eb6e7a03c.

packages/core/contracts/Margin.sol
EdNoepel: Okay; I will deprecate `Local.claimable`, move that into it's own collection in `Margin` contract, and _not_ checkpoint it.  This implies we'll still need separate **updateBalance** for `Market.claimFee` and (temporarily) `Market.claimExposure`, and **updateCheckpoint** for processing orders.

EdNoepel: Should `Margin.claimFee` push claimable tokens, or move the balance to the user's cross-margin balance?

kbrizzle: Let's `claimFee` push claimable tokens like we do now in the Market. (we may need to hook that up to the oracle settlement callback flow as well to get those tests passing).

packages/core/contracts/Market.sol
EdNoepel: This no longer applies since `close` method cannot manipulate collateral, though I left the check in-place, as it remains in `v2.4` branch _InvariantLib_.

packages/core/contracts/Margin.sol
EdNoepel: Merged the branch referenced above.

packages/core/contracts/Market.sol
EdNoepel: New implementation makes a call to `Margin.updateClaimable` here to claim global fees, as `Local.claimable` has been deprecated.

packages/vault/contracts/Vault.sol
EdNoepel: cleaned that up in e73b6e95a4f7d799f95d0aafd65115eeefb20be7

kbrizzle (approved):

prateekdefi (approved):

PR #508, opened 2024-12-02
Split `maxSettlementFee` into `maxSyncFee` and `maxAsyncFee`

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1982/[feature]-split-maxsettlementfee

Migration Note: 
- Reset Oracle parameters and set new parameters


EdNoepel (commented):
Please add deployment notes to PR description, stating that all oracle parameters must be reset in the deployment.  Otherwise, the first 24 bits of `maxSyncFee` would be read from where `maxOracleFee` was stored in v2.3.

packages/oracle/contracts/keeper/types/PriceResponse.sol
```diff
@@ -69,15 +69,17 @@ library PriceResponseLib {
...
-    function applyFeeMaximum(PriceResponse memory self, UFixed6 maxSettlementFee, uint256 callbacks) internal pure {
-        UFixed6 totalSettlementFee = settlementFee(self, callbacks);
-        if (totalSettlementFee.gt(maxSettlementFee)) {
-            self.syncFee = self.syncFee.muldiv(maxSettlementFee, totalSettlementFee);
-            self.asyncFee = self.asyncFee.muldiv(maxSettlementFee, totalSettlementFee);
-        }
+    /// @param maxSyncFee The max sync fee
+    /// @param maxAsyncFee The max async fee
+    /// @param callbacks The number of settlement callbacks to be made
+    function applyFeeMaximum(PriceResponse memory self, UFixed6 maxSyncFee, UFixed6 maxAsyncFee, uint256 callbacks) internal pure {
```
kbrizzle, line 77: instead of
```
if (self.syncFee.gt(maxSyncFee))
    self.syncFee = maxSyncFee;
```
it's a bit cleaner to write these caps as:
```
self.syncFee = self.syncFee.min(maxSyncFee);
```

packages/oracle/contracts/keeper/types/PriceResponse.sol
```diff
@@ -69,15 +69,17 @@ library PriceResponseLib {
...
+        if (self.syncFee.gt(maxSyncFee))
+            self.syncFee = maxSyncFee;
+        UFixed6 asyncFee = self.asyncFee.mul(UFixed6Lib.from(callbacks));
```
kbrizzle, line 80: The async fee maximum will cap the individual async fees charges (so that if we have tons of orders in a single version we're not capping the total amount charged).

So no need to multiply this by the callback amount, that won't be needed anymore.

packages/oracle/contracts/keeper/KeeperOracle.sol
```diff
@@ -207,8 +207,8 @@ contract KeeperOracle is IKeeperOracle, Instance {
         priceResponse.asyncFee = UFixed6Lib.from(factory.settlementGasOracle().cost(0), true);
         priceResponse.oracleFee = keeperOracleParameter.oracleFee;
```
kbrizzle, line 208: actually now that `applyFeeMaximum` is so simple, maybe we can just apply the `.min(oracleParameter....)` directly on these above and delete the helper function altogether.

prateekdefi: Removed helper method in [aac4ffe](https://github.com/equilibria-xyz/perennial-v2/pull/508/commits/aac4ffed95e64e86d983bd42b8ea409805355e2a)

kbrizzle (approved):

PR #509, opened 2024-12-03
Consolidate operator settings to only marketFactory

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1979/[cleanup]-consolidate-operator-settings-to-only-marketfactory

Migration Note:
- Ensure that multiInvoker and vault users are notified to approve marketFactory for authorization.
- Update frontend to approve marketFactory for all multiInvoker and vault transactions.


EdNoepel (commented):
My concern here is migration.  During deployment, we can't iterate through the mapping, and even if we could, _owner_ cannot (and should not) update operators in _MarketFactory_ on behalf of users.  So we'll need to inform _MultiInvoker_ and _Vault_ users they they must interact with _MarketFactory_ to re-establish their authorizations.

Should add such a note to the PR description to record this in the deployment plan.

kbrizzle:
Yeah UX here will be to do a one-time approval popup (which should just come up automatically) if we need approvals given the new state.

Let's make a note of this though so we're aware on the frontend interface side.

kbrizzle (approved):

EdNoepel (approved):
GTG

PR #511, opened 2024-12-05
Permissioned vaults

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1877/[feature]-permissioned-vaults


packages/vault/contracts/Vault.sol
```diff
@@ -270,6 +274,14 @@ contract Vault is IVault, Instance {
         _saveContext(context, account);
     }
 
+    /// @notice Updates the allowed account status for a given account
+    /// @param account The account to update
+    /// @param allowed_ The new allowed status
+    function updateAllowedAccount(address account, bool allowed_) public onlyOwner {
```
kbrizzle, line 280: naming:
- function should be called `updateAllowed`
- parameter should be called `newAllowed`.

packages/vault/contracts/Vault.sol
```diff
@@ -230,7 +234,7 @@ contract Vault is IVault, Instance {
 
     /// @notice Syncs `account`'s state up to current
     /// @param account The account that should be synced
-    function settle(address account) public whenNotPaused {
+    function settle(address account) public whenNotPaused onlyAllowed {
```
kbrizzle, line 237: I think we only need `onlyAllowed` on `update()`. The goal is prevent non-allowed users from being able to deploy capital to these Vaults, so `settle()` / `rebalance()` should be ok to leave public.

packages/vault/contracts/Vault.sol
```diff
@@ -71,6 +74,7 @@ contract Vault is IVault, Instance {
         _name = name_;
         _register(initialMarket);
         _updateParameter(VaultParameter(initialDeposit, UFixed6Lib.ZERO));
+        allowed[msg.sender] = true;
```
kbrizzle, line 77: Is this required for the initial deposit logic? This leaves the original sender (often a deployment EOA) access to the vault even if that was not the intention.

Can we do some sort of `updateAllowedAccount()` on / off flow in the `MarketFactory.initializer()` instead?

packages/vault/contracts/Vault.sol
```diff
@@ -270,6 +274,14 @@ contract Vault is IVault, Instance {
         _saveContext(context, account);
     }
 
+    /// @notice Updates the allowed account status for a given account
+    /// @param account The account to update
+    /// @param allowed_ The new allowed status
+    function updateAllowedAccount(address account, bool allowed_) public onlyOwner {
+        allowed[account] = allowed_;
+        emit AllowedAccountUpdated(account, allowed_);
```
kbrizzle, line 282: naming: `AllowedUpdate`

packages/vault/test/integration/vault/Vault.test.ts
```diff
@@ -639,6 +642,29 @@ describe('Vault', () => {
     })
   })
 
+  describe('#updateAllowedAccount', () => {
```
kbrizzle, line 645: It's cleaner if the `#updateAllowedAccount` contains tests for updating the settings / verifying the state was properly updated, then adding addition `revert` clauses to the `#update` to test the functionality.

packages/vault/test/integration/vault/Vault.test.ts
```diff
@@ -639,6 +642,29 @@ describe('Vault', () => {
...
+      await expect(vault.connect(user).update(user.address, parse6decimal('10'), 0, 0)).to.be.revertedWithCustomError(
+        vault,
+        'VaultNotAllowedError',
+      )
+
+      // allow user to interact with the vault
+      await vault.connect(owner).updateAllowedAccount(user.address, true)
+
+      // should update allowed
+      expect(await vault.allowed(user.address)).to.be.true
```
kbrizzle, line 661: we want to test the `address(0)` case as well.

packages/vault/contracts/Vault.sol
prateekdefi: If an EOA deploys the vault directly, there can be inflation attack and also `onlyOwner` check would fail as EOA will be set as factory contract and `factory().owner()` will fail.

If vault factory creates the vault then vault factory contract would be allowed, which I think is safe and also factory owner can disable the factory as allowed. What do u think @kbrizzle

Also, I think it is not possible to create on/off flow in `VaultFactory.create()` because only owner can update the allowed and vault factory is not the owner.

packages/vault/contracts/Vault.sol
EdNoepel: Curious why we want to switch to present tense for this event; `AllowedUpdated` seems more appropriate.

EdNoepel (commented):
Seems fine other than the `initialize` behavior.

packages/vault/contracts/interfaces/IVault.sol
```diff
@@ -39,6 +39,7 @@ interface IVault is IInstance {
     event MarketUpdated(uint256 indexed marketId, UFixed6 newWeight, UFixed6 newLeverage);
     event ParameterUpdated(VaultParameter newParameter);
     event Updated(address indexed sender, address indexed account, uint256 version, UFixed6 depositAssets, UFixed6 redeemShares, UFixed6 claimAssets);
+    event AllowedUpdate(address indexed account, bool newAllowed);
```
kbrizzle, line 42: Sorry I had a typo, this should be `AllowedUpdated` --

In general all events like this should be `VariableNameUpdated` for single-variable update events.

packages/vault/contracts/interfaces/IVault.sol
```diff
@@ -99,4 +102,6 @@ interface IVault is IInstance {
     function updateLeverage(uint256 marketId, UFixed6 newLeverage) external;
     function updateWeights(UFixed6[] calldata newWeights) external;
     function updateParameter(VaultParameter memory newParameter) external;
+    function allowed(address account) external view returns (bool);
```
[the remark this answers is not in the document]
kbrizzle: Can we move these in one pass in a standalone PR? would be cleaner to track.

Or maybe a PR per package or something if we want to break it up.

packages/vault/contracts/Vault.sol
```diff
@@ -54,6 +54,9 @@ contract Vault is IVault, Instance {
     /// @dev DEPRECATED SLOT -- previously the mappings
     bytes32 private __unused0__;
 
+    /// @dev Mapping to track allowed accounts
+    mapping(address => bool) public allowed;
```
kbrizzle, line 58: note: this is ok only because `coordinator` has been added in v2.4 and is not currently live in v2.3.

prateekdefi: okay, got it

packages/vault/contracts/Vault.sol
EdNoepel: ⬆ @kbrizzle

packages/vault/contracts/interfaces/IVault.sol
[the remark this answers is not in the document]
EdNoepel: Doesn't matter to me, but v2.4 branch already has some documented the new way:
https://github.com/equilibria-xyz/perennial-v2/blob/v2.4/packages/core/contracts/interfaces/IMarket.sol#L75
https://github.com/equilibria-xyz/perennial-v2/blob/v2.4/packages/core/contracts/interfaces/IMarket.sol#L108

packages/vault/contracts/Vault.sol
prateekdefi: I have made the update to past tense as well

packages/vault/contracts/interfaces/IVault.sol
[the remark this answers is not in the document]
kbrizzle: I see, yeah then feel free to add these whenever makes most sense.

kbrizzle (approved):

PR #513, opened 2024-12-17
prevent close order from executing if account has no position

EdNoepel:
Seems there is a use case where users submit a trigger order to close their position alongside an intent for opening a position.  When this happens, the trigger order becomes immediately executable.  When the trigger order executes as a no-op update, the UX is that their trigger order disappeared.

To work around this behavior, update logic such that trigger orders to close are not executable if user has no position.

## TODO
- [x] Self-review
- [ ] Merge this to v2.4 branch after approval

### DIFF
--- a/packages/perennial-order/contracts/types/TriggerOrder.sol
+++ b/packages/perennial-order/contracts/types/TriggerOrder.sol
@@ -38,13 +38,24 @@ library TriggerOrderLib {
 
     /// @notice Determines whether the trigger order is fillable at the latest price
     /// @param self Trigger order
-    /// @param latestVersion Latest oracle version
+    /// @param market Market for which the order is intended to be executed
+    /// @param account Market participant
     /// @return Whether the trigger order is fillable
-    function canExecute(TriggerOrder memory self, OracleVersion memory latestVersion) internal pure returns (bool) {
+    function canExecute(TriggerOrder memory self, IMarket market, address account) internal view returns (bool) {
+        OracleVersion memory latestVersion = market.oracle().latest();
+
         if (!latestVersion.valid) return false;
-        if (self.comparison == 1) return latestVersion.price.gte(self.price);
-        if (self.comparison == -1) return latestVersion.price.lte(self.price);
-        return false;
+        if (self.comparison == 1 && latestVersion.price.lt(self.price)) return false;
+        if (self.comparison == -1 && latestVersion.price.gt(self.price)) return false;
+        if (self.comparison != 1 && self.comparison != -1) revert TriggerOrderInvalidError();
+
+        if (self.delta.eq(MAGIC_VALUE_CLOSE_POSITION)) {
+            Position memory position = market.positions(account);
+            // prevent execution if the position was not opened or is already closed
+            if (position.empty()) return false;
+        }
+
+        return true;
     }
 
     /// @notice Applies the order to the user's position and updates the market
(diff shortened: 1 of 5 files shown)


packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -38,13 +38,24 @@ library TriggerOrderLib {
...
-        if (self.comparison == -1) return latestVersion.price.lte(self.price);
-        return false;
+        if (self.comparison == 1 && latestVersion.price.lt(self.price)) return false;
+        if (self.comparison == -1 && latestVersion.price.gt(self.price)) return false;
+        if (self.comparison != 1 && self.comparison != -1) revert TriggerOrderInvalidError();
+
+        if (self.delta.eq(MAGIC_VALUE_CLOSE_POSITION)) {
+            Position memory position = market.positions(account);
+            // prevent execution if the position was not opened or is already closed
+            if (position.empty()) return false;
```
EdNoepel, line 55: Wasn't sure if we wanted to get fancy and match the position with the order's side. This implementation seems most maintainable and gas efficient.

packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -38,13 +38,24 @@ library TriggerOrderLib {
...
         if (!latestVersion.valid) return false;
-        if (self.comparison == 1) return latestVersion.price.gte(self.price);
-        if (self.comparison == -1) return latestVersion.price.lte(self.price);
-        return false;
+        if (self.comparison == 1 && latestVersion.price.lt(self.price)) return false;
+        if (self.comparison == -1 && latestVersion.price.gt(self.price)) return false;
+        if (self.comparison != 1 && self.comparison != -1) revert TriggerOrderInvalidError();
+
+        if (self.delta.eq(MAGIC_VALUE_CLOSE_POSITION)) {
+            Position memory position = market.positions(account);
```
EdNoepel, line 53: This extra storage read means close orders will be a little more expensive than non-close orders.  The check happens inside gas measurement for `Manager.executeOrder`, so shouldn't require keeper config adjustment.

packages/perennial-order/contracts/types/TriggerOrder.sol
```diff
@@ -38,13 +38,24 @@ library TriggerOrderLib {
 
     /// @notice Determines whether the trigger order is fillable at the latest price
     /// @param self Trigger order
-    /// @param latestVersion Latest oracle version
+    /// @param market Market for which the order is intended to be executed
+    /// @param account Market participant
     /// @return Whether the trigger order is fillable
-    function canExecute(TriggerOrder memory self, OracleVersion memory latestVersion) internal pure returns (bool) {
+    function canExecute(TriggerOrder memory self, IMarket market, address account) internal view returns (bool) {
```
kbrizzle, line 44: This change as-is is a bit too complex for a hot fix, is there any way we can pair this down?

My idea would be to revert the latestOracle change and pass in the local position here instead. (making both calls to market prior to this function).

Then you could add a simple guard condition instead of change the logic here:
```
if (self.delta.eq(MAGIC_VALUE_CLOSE_POSITION) && position.empty()) return false;
```

kbrizzle: Another issue is that this code path does not settle the market prior to this call, so directly querying position won't have an up to date result.

Also This won't take into account pending positions (there is already a full close in flight so, a zero-value close would be a max close).

PR #514, opened 2024-12-17
TriggerOrder close order logic changes

EdNoepel:
There is a use case where users submit a trigger order to close their position alongside an intent for opening a position.  When this happens, the trigger order becomes immediately executable.  When the trigger order executes as a no-op update, the UX is that their trigger order disappeared.

To work around this behavior, update logic such that trigger orders to close are not executable if user has no position.

This PR is a follow-up to #513.  Differences:
- This moves logic out of `TriggerOrder` type and into `Manager`, as requested in the other PR.
- This also checks pending positions.

Note that I do not necessarily agree with the first change. Passing _market_ and _account_ into the the method seems better for maintainability.  I recommend we take that approach when merging into v2.4 branch, unless we wish to abandon this behavior and implement a _trigger order dependency on intent execution_ feature instead.

## TODO
- [x] Self-review
- [ ] Merge this to v2.4 branch after approval

### DIFF
--- a/packages/perennial-order/contracts/Manager.sol
+++ b/packages/perennial-order/contracts/Manager.sol
@@ -131,7 +131,7 @@ abstract contract Manager is IManager, Kept {
         order = _orders[market][account][orderId].read();
         // prevent calling canExecute on a spent or empty order
         if (order.isSpent || order.isEmpty()) revert ManagerInvalidOrderNonceError();
-        canExecute = order.canExecute(market.oracle().latest());
+        canExecute = order.canExecute(market, account);
     }
 
     /// @inheritdoc IManager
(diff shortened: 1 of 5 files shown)


packages/perennial-order/contracts/Manager.sol
```diff
@@ -131,7 +132,16 @@ abstract contract Manager is IManager, Kept {
...
         if (order.isSpent || order.isEmpty()) revert ManagerInvalidOrderNonceError();
-        canExecute = order.canExecute(market.oracle().latest());
+        Position memory position;
+        if (order.delta.eq(TriggerOrderLib.MAGIC_VALUE_CLOSE_POSITION)) {
+            position = market.positions(account);
+            Order memory pending = market.pendings(account);
+            position.update(pending);
+        } else {
+            position = Position(0, UFixed6Lib.ZERO, UFixed6Lib.ZERO, UFixed6Lib.ZERO);
+        }
```
EdNoepel, line 142: This feels awkward, but avoids encumbering every order execution from reading three additional storage slots.

PR #487, opened 2024-10-27
Add synthetic orderbook

kbrizzle:
#### Migration Note
- Removes exposure
  - Need to set adiabatic fees to zero and settle exposure ahead of migration
- Changes storage layout of `Version` and `Guarantee`
  - Need to settle-only / settle all markets & vaults

PR #523, opened 2025-01-05
Split takerPos / takerNeg in Guarantee

kbrizzle:
Based off of: https://github.com/equilibria-xyz/perennial-v2/pull/487.

Enables invalidating only the non-Intent portion of an Order.

#### Migration Note
- Must settle up all outstanding guarantees prior to migration


EdNoepel (approved):
LGTM.

Regarding deployment: once markets are settle-only mode, settlement of pending orders should empty out the guarantee collections in _Market_, per the deployment note in the PR description.  For intents which remain offchain, after contracts are upgraded, seems the updated `GuaranteeLib.from` should correctly populate the new _Guarantee_.

PR #525, opened 2025-01-05
Vault structure refactor

kbrizzle:
Reorganizes how the Vault's strategy is called so that it is cleaner to extend the base Vault contract with additional strategies.


EdNoepel (approved):
LGTM.

PR #527, opened 2025-01-06
Add orderReferral field to Guarantee

kbrizzle:
Renames `referral` to `solverReferral` and adds a corresponding `orderReferral` field to both the local and global `Guarantee`.

This allows us to track the total referral applicable to the intent for future invalidation tracking.


EdNoepel (approved):
LGTM

PR #528, opened 2025-01-06
Always execute intent orders

kbrizzle:
Changes invalidation flow to only invalidate the non-intent (Guarantee) portion of the Order since we do not need a fresh price to open an Intent-based order.

PR #529, opened 2025-01-06
Skip price request for intents

kbrizzle:
Intents no longer need to request a price on settlement (only on open for the margin check) due to: https://github.com/equilibria-xyz/perennial-v2/pull/528.


packages/core/contracts/Market.sol
```diff
@@ -694,8 +689,8 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         // apply referrer
         _processReferrer(updateContext, newOrder, newGuarantee, orderReferrer, guaranteeReferrer);
 
-        // request version, only request new price on position change
-        if (!newOrder.isEmpty()) oracle.request(IMarket(this), context.account);
+        // request version, only request new price on non-empty market order
+        if (!newOrder.isEmpty() && newGuarantee.isEmpty()) oracle.request(IMarket(this), context.account);
```
EdNoepel, line 693: Not sure I understand this logic.  So the price request will only be made if new order is not empty, but the guarantee is empty.  That means if the new order is not empty, but guarantee is also not empty, no request should be made?

EdNoepel (commented):
Wasn't stated in the ticket, but seems eliminating the settlement fee from intent orders was the goal of skipping the price request.  Question and comment inline.

packages/core/test/unit/market/Market.test.ts
```diff
@@ -23024,6 +23005,268 @@ describe('Market', () => {
...
+            })
+            expectOrderEq(await market.pendingOrders(user.address, 1), {
+              ...DEFAULT_ORDER,
+              timestamp: ORACLE_VERSION_3.timestamp,
+              orders: 1,
+              longPos: POSITION.div(2),
+              takerReferral: POSITION.div(2).mul(2).div(10),
+              collateral: COLLATERAL,
+            })
+            expectCheckpointEq(await market.checkpoints(user.address, ORACLE_VERSION_4.timestamp), {
```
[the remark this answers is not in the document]
kbrizzle: Added in https://github.com/equilibria-xyz/perennial-v2/pull/529/commits/4993824013a458ecbcde0a6b7841d9b64e30a132.

packages/core/contracts/Market.sol
kbrizzle: Right, this is because a Guarantee is a subset of an order, so any non-empty update will create an Order, but only non-empty intents will create a Guarantee.

- order empty / guarantee empty -> no-op order (don't need a price)
- order non-empty / guarantee empty -> market order (need a price)
- order non-empty / guarantee non-empty -> intent order (don't need a price)

kbrizzle:
> Wasn't stated in the ticket, but seems eliminating the settlement fee from intent orders was the goal of skipping the price request. Question and comment inline.

Right yeah, the goal of this falls under general fee reduction which is accomplished by removing the need to charge the settlement fee since we're not requesting a price.

EdNoepel (approved):
LGTM

PR #536, opened 2025-01-10
Market taker update through signed message

EdNoepel:
# Purpose
Establishes a foundation by which users may change their position through a signed message without using an intent order.  The message implemented in this PR limits users to changing their taker position without collateral change.

# Scope
Explicitly out-of-scope for this PR is relaying of this message.

# Status
- [x] Check coverage of affected contracts
- [x] Self-review
- [x] Confirm CI runs clean

### DIFF
--- a/packages/core/contracts/Market.sol
+++ b/packages/core/contracts/Market.sol
@@ -28,7 +28,6 @@ import { InvariantLib } from "./libs/InvariantLib.sol";
 import { MagicValueLib } from "./libs/MagicValueLib.sol";
 import { VersionAccumulationResponse, VersionLib } from "./libs/VersionLib.sol";
 import { CheckpointAccumulationResponse, CheckpointLib } from "./libs/CheckpointLib.sol";
-import "hardhat/console.sol";
 
 /// @title Market
 /// @notice Manages logic and state for a single market.
--- a/packages/core/contracts/Market.sol
+++ b/packages/core/contracts/Market.sol
@@ -21,7 +21,7 @@ import { Order, OrderLib, OrderStorageGlobal, OrderStorageLocal } from "./types/
 import { Guarantee, GuaranteeLib, GuaranteeStorageGlobal, GuaranteeStorageLocal } from "./types/Guarantee.sol";
 import { Checkpoint, CheckpointStorage } from "./types/Checkpoint.sol";
 import { Intent } from "./types/Intent.sol";
-import { MarketUpdateTaker } from "./types/MarketUpdateTaker.sol";
+import { Take } from "./types/Take.sol";
 import { OracleVersion } from "./types/OracleVersion.sol";
 import { OracleReceipt } from "./types/OracleReceipt.sol";
 import { InvariantLib } from "./libs/InvariantLib.sol";
@@ -176,8 +176,8 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     /// @notice Updates the account's taker position without collateral change
     /// @param update_ Message requesting change in user's taker position
     /// @param signature Taker's signature
-    function update(MarketUpdateTaker calldata update_, bytes memory signature) external {
-        verifier.verifyMarketUpdateTaker(update_, signature);
+    function update(Take calldata update_, bytes memory signature) external {
+        verifier.verifyTake(update_, signature);
         (Context memory context, UpdateContext memory updateContext) =
             _loadForUpdate(update_.common.account, update_.common.account, update_.referrer, address(0), UFixed6Lib.ZERO, UFixed6Lib.ZERO);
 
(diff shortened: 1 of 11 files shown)


packages/core/test/integration/core/happyPath.test.ts
```diff
@@ -39,8 +45,14 @@ export const TIMESTAMP_5 = 1631118731
 export const PRICE_0 = parse6decimal('113.882975')
 export const PRICE_1 = parse6decimal('113.796498')
 export const PRICE_2 = parse6decimal('115.046259')
+export const PRICE_3 = parse6decimal('116.284753')
 export const PRICE_4 = parse6decimal('117.462552')
 
+const COMMON_PROTOTYPE = '(address,address,address,uint256,uint256,uint256)'
+const MARKET_UPDATE_ABSOLUTE_PROTOTYPE = 'update(address,uint256,uint256,uint256,int256,bool)'
+const MARKET_UPDATE_DELTA_PROTOTYPE = 'update(address,int256,int256,address)'
+const MARKET_UPDATE_TAKER_PROTOTYPE = `update((int256,address,${COMMON_PROTOTYPE}),bytes)`
```
EdNoepel, line 54: Not sure how we feel about this.  Seemingly makes it easier to manage the growing list of `Market.update` overrides.  At a minimum, I'd like to keep `COMMON_PROTOTYPE` for reuse with intent tests.

packages/core/contracts/Market.sol
```diff
@@ -172,6 +173,30 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         ); // signer
     }
 
+    /// @notice Updates the account's taker position without collateral change
+    /// @param update_ Message requesting change in user's taker position
+    /// @param signature Taker's signature
+    function update(MarketUpdateTaker calldata update_, bytes memory signature) external {
+        verifier.verifyMarketUpdateTaker(update_, signature);
+        (Context memory context, UpdateContext memory updateContext) =
+            _loadForUpdate(update_.common.account, update_.common.account, update_.referrer, address(0), UFixed6Lib.ZERO, UFixed6Lib.ZERO);
```
EdNoepel, line 182: I couldn't simply call `update(address,uint256,uint256,uint256,address)` here because it zeros out the signer.  Considered refactoring into some private `_updateMarket` method, but was concerned about moving logic to bottom of file and causing confusion with existing common `_update` private method.

kbrizzle: I think we'll need go the `_updateMarket` route due to code size, especially with the other intent with signature update we're adding as well.

The flows here should be pretty intuitive:
- `update(intent) / update(intent, signature) -> _updateIntent() -> _update()`
- `update(delta) / update(delta, signate) -> _updateMarket() -> _update()`

kbrizzle, line 182: second parameter should be `common.signer`, we should add a test to check this case (an update signed by a delegated signer of the account).

packages/core/test/integration/core/happyPath.test.ts
kbrizzle: Hmm yeah this probably makes sense. My IDE autocompletes the function signature, whereas it wouldn't be able to fine these, but these are more legible.

packages/core/test/unit/verifier/Verifier.test.ts
```diff
@@ -349,6 +350,226 @@ describe('Verifier', () => {
...
+        nonce: 0,
+        group: 0,
+        expiry: constants.MaxUint256,
+      },
+    }
+
+    it('should verify MarketUpdateTaker message', async () => {
+      const marketUpdateTaker = {
+        ...DEFAULT_MARKET_UPDATE_TAKER,
+        referrer: constants.AddressZero, //.address,
```
kbrizzle, line 370: nit: comment

packages/core/contracts/types/MarketUpdateTaker.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+
+import { UFixed6 } from "@equilibria/root/number/types/UFixed6.sol";
+import { Fixed6 } from "@equilibria/root/number/types/Fixed6.sol";
+import { Common, CommonLib } from "@equilibria/root/verifier/types/Common.sol";
+
+/// @notice Market update which modifies the taker's position without collateral change
+struct MarketUpdateTaker {
```
kbrizzle, line 9: Since this struct is external to the integrators, it'd be nice to have a simple name for it (even if less precise).

For intents we have
- `Intent` taker-signed
- `Fill` taker-solver-signed

For this maybe:
- `Take` ?

Lemme know if you have other ideas.

packages/core/contracts/Market.sol
EdNoepel: Done in c829383a94ae05ac777e7dedbf8d2801f626e417.

EdNoepel: Oof 😬 thanks for catching.  Fixed and added a test to ensure the revert happens in c829383a94ae05ac777e7dedbf8d2801f626e417.

packages/core/test/unit/verifier/Verifier.test.ts
EdNoepel: Cleaned up in edee4d2544ccba32a82d5b0dd4524987c982c655.

kbrizzle (approved):

PR #537, opened 2025-01-12
Solver Vault

kbrizzle:


packages/vault/test/integration/vault/SolverVault.test.ts
```diff
@@ -0,0 +1,1622 @@
...
+  }
+
+  async function totalCollateralInVault() {
+    return (await collateralInVault())
+      .add(await btcCollateralInVault())
+      .mul(1e12)
+      .add(await asset.balanceOf(vault.address))
+  }
+
+  async function currentPositionGlobal(market: IMarket) {
```
EdNoepel, line 142: This doesn't appear to be used anywhere.

packages/vault/test/integration/vault/SolverVault.test.ts
EdNoepel, line 1: File seems to be copypasted from `MakerVault.test.ts` without refactoring to share logic.  Would be nice if we had a `setupHelpers.ts` to create the fixture shared between both test suites.

EdNoepel (commented):
No issues with the solidity code.  Refactoring integration tests can be deferred, but need to determine why SolverVault integration tests cannot be run after MakerVault integration tests.

EdNoepel (approved):
LGTM

packages/vault/test/integration/vault/SolverVault.test.ts
kbrizzle: We also now have many "Vault" tests in the `MakerVault` suite, while `SolverVault` only contains tests for its implementation, so these could also be refactored for clarity.

PR #538, opened 2025-01-12
Add invalidation semaphore

kbrizzle:
Adds a semaphore to the local `Order` to track the quantity of pending AMM orders.
- This will be used to validate when it is safe to allow a non-reduce-only intent order to be placed.

#### Migration Note
- Must settle up all outstanding local orders


EdNoepel (approved):
Change seems fine, though I didn't see any test which `add`ed multiple orders to increment the counter past 1.

More concerned about the existing `protection` semaphore.  Curious about why it's okay to limit max value to 8 (1 byte).

PR #542, opened 2025-01-17
Add zero-cross for Intents

kbrizzle:
Enables crossing zero in delta update methods (for intent trading).

Adds testing around invariant updates to support this feature.

### DIFF
--- a/packages/core/test/unit/market/Market.test.ts
+++ b/packages/core/test/unit/market/Market.test.ts
@@ -20927,6 +20927,385 @@ describe('Market', () => {
         })
       })
 
+      context('zero crossing', async () => {
+        beforeEach(async () => {
+          dsu.transferFrom.whenCalledWith(user.address, market.address, COLLATERAL.mul(1e12)).returns(true)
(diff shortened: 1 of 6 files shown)


packages/core/test/unit/market/Market.test.ts
```diff
@@ -407,6 +409,19 @@ async function deposit(market: Market, amount: BigNumber, account: SignerWithAdd
     )
 }
 
+async function getOrderProcessingEvents(
```
EdNoepel, line 412: Where is this used?  Could the one from `fees.test.ts` have been moved into _common_ and reused?

EdNoepel (approved):
Looks good.  As far as integration testing, should something in `happyPath.test.ts` exercise a zero-cross?

packages/core/test/unit/market/Market.test.ts
kbrizzle: I use it heavily while debugging so I thought I'd add it in here instead of copy-pasting it every time.

kbrizzle:
I added a follow up task to add integration tests for zero-crosses.

PR #543, opened 2025-01-17
Default signer to sender

kbrizzle:
Defaults the `signer` value to the `msg.sender` in cases where there is no signature in the update method.

This allows delegated signers to send direct transactions to the Market update functions (proving their source) as an alternative to the "with siganture" counterparts.


packages/core/contracts/Market.sol
```diff
@@ -209,11 +210,20 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     /// @notice Updates the account's taker position without collateral change
     /// @param take Message requesting change in user's taker position
     /// @param signature Signature of taker or authorized signer
-    function update(Take calldata take, bytes memory signature) external {
+    function update(Take calldata take, bytes memory signature) external nonReentrant whenNotPaused {
         verifier.verifyTake(take, signature);
         _updateMarket(take.common.account, take.common.signer, Fixed6Lib.ZERO, take.amount, Fixed6Lib.ZERO, take.referrer);
     }
 
+    /// @notice Updates the account's position and collateral
```
EdNoepel, line 218:
```suggestion
    /// @notice Updates the account's position
```

EdNoepel (commented):
I believe I understand the motivation behind this, but an auditor would likely flag this as escalation of privileges.  It feels awkward to break limitations of the _signer_ role by allowing them to call a few specific _update_ methods.  

Is this worth the goal defined in PE-2139?

EdNoepel (approved):
Recommend updating the comment on line 39 of `MarketFactory.sol` to define the role.

Should consider renaming _signer_ in 2.5, possibly to _coordinator_ for consistency with _Vault_ nomenclature.

kbrizzle:
Added some more clarity on the roles in `MarketFactory`

PR #545, opened 2025-01-18
Trigger Orders Manager contract to compensate keepers and pay interface fees from collateral accounts rather than market

EdNoepel:
To reduce price commitments, have _Manager_ contract use collateral accounts for compensating keepers and paying interface fees.  Since `IAccount.withdraw` cannot be used for this purposes, `IController` has been extended with a `chargeFee` method only callable from an operator.  This facility also wraps USDC if necessary.  Since trigger order `Manager` already needs to be a manager to interact with markets on behalf of the user, no new entitlements are required.  

## Deployment notes
Trigger order users must create and fund their collateral account prior to using the new version.

### DIFF
--- a/packages/periphery/README.md
+++ b/packages/periphery/README.md
@@ -84,7 +84,7 @@ For executing orders and handling signed messages, keepers are compensated from 
 ### Usage
 
 #### Users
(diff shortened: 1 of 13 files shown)


kbrizzle (approved):

kbrizzle:
Very clean implementation!

PR #546, opened 2025-01-20
Add rebalance to SolverVault

kbrizzle:
Adds a rebalance function to the SolverVault which allows the `coordinator` to move collateral from one registered market to another.


EdNoepel (commented):
Seems the issue with the custom errors was resolved before I could comment on it.  No issues otherwise; just waiting on CI.

EdNoepel (approved):
LGTM

PR #549, opened 2025-01-21
Add second VaultFactory to MultiInvoker

kbrizzle:
To support a separate isntance for the new `SolverVault`.


EdNoepel (approved):
CI issues should be resolved when this is merged with `v2.4-mini`.

PR #550, opened 2025-01-21
[Code Size] v2.4 mini

kbrizzle:
#### Order
- Makes external
  - `LocalOrderLib.store()`
  - `GlobalOrderLib.store()`
- Logic cleanup
   - `Order.from()`
#### Guarantee
- Makes external
  - `LocalGuaranteeLib.store()`
  - `GlobalGuaranteeLib.store()`
#### MagicValueLib
- Makes internal
  - `MagicValueLib.process()`
- Removes redundant clause [here](https://github.com/equilibria-xyz/perennial-v2/pull/550/files#diff-0a101158e345c3ee4438c191c98ab25025e9421ad80c9a6549fb9af660ea34b2L57).
  - If zero-cross, a zero currentPosition is returned
  - else .min(currentPosition) returns a zero currentPosition

Saves 948 bytes.

### DIFF
--- a/packages/core/contracts/libs/MagicValueLib.sol
+++ b/packages/core/contracts/libs/MagicValueLib.sol
@@ -20,7 +20,7 @@ library MagicValueLib {
         UFixed6 newMaker,
         UFixed6 newLong,
         UFixed6 newShort
-    ) external pure returns (Fixed6, UFixed6, UFixed6, UFixed6) {
+    ) internal pure returns (Fixed6, UFixed6, UFixed6, UFixed6) {
         return (
             _processCollateralMagicValue(context, collateral),
             _processPositionMagicValue(context, updateContext.currentPositionLocal.maker, newMaker),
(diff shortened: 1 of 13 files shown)


packages/core/contracts/libs/MagicValueLib.sol
```diff
@@ -54,7 +54,6 @@ library MagicValueLib {
     ) private pure returns (UFixed6) {
         if (newPosition.eq(MAGIC_VALUE_UNCHANGED_POSITION)) return currentPosition;
         if (newPosition.eq(MAGIC_VALUE_FULLY_CLOSED_POSITION)) {
-            if (currentPosition.isZero()) return UFixed6Lib.ZERO;           // side is empty
```
EdNoepel, line 57: Confirmed logic on line below handles this case.

packages/core/contracts/types/Order.sol
```diff
@@ -148,35 +148,46 @@ library OrderLib {
...
-            Fixed6Lib.from(newShort).sub(Fixed6Lib.from(position.short))
+            _change(position.maker, newMaker),
+            _change(position.long, newLong),
+            _change(position.short, newShort)
         );
 
-        // create the new order
-        newOrder = Order(
-            timestamp,
-            0,
```
EdNoepel, line 159: We can trust order count will default to 0.

packages/core/contracts/types/Order.sol
```diff
@@ -148,35 +148,46 @@ library OrderLib {
...
-            collateral,
-            makerAmount.max(Fixed6Lib.ZERO).abs(),
-            makerAmount.min(Fixed6Lib.ZERO).abs(),
-            longAmount.max(Fixed6Lib.ZERO).abs(),
-            longAmount.min(Fixed6Lib.ZERO).abs(),
-            shortAmount.max(Fixed6Lib.ZERO).abs(),
-            shortAmount.min(Fixed6Lib.ZERO).abs(),
-            protect ? 1 : 0,
-            0,
```
EdNoepel, line 168: These two lines are handled on lines 169-173.

PR #552, opened 2025-01-22
Move Local.claimable to Margin contract...

EdNoepel:
...and create a facility in `Margin` contract for claiming fees.

### DIFF
--- a/packages/core/contracts/InsuranceFund.sol
+++ b/packages/core/contracts/InsuranceFund.sol
@@ -4,11 +4,12 @@ pragma solidity 0.8.24;
 import { Ownable } from "@equilibria/root/attribute/Ownable.sol";
 import { Token18 } from "@equilibria/root/token/types/Token18.sol";
 import { Fixed6, Fixed6Lib } from "@equilibria/root/number/types/Fixed6.sol";
+import { UFixed6, UFixed6Lib } from "@equilibria/root/number/types/UFixed6.sol";
 import { IFactory } from "@equilibria/root/attribute/interfaces/IFactory.sol";
 import { IInstance } from "@equilibria/root/attribute/interfaces/IInstance.sol";
 
 import { IInsuranceFund } from "./interfaces/IInsuranceFund.sol";
-import { IMarket } from "./interfaces/IMarket.sol";
+import { IMarket, IMargin } from "./interfaces/IMarket.sol";
 
 /// @title InsuranceFund
 /// @notice This contract manages the protocol fee and shortfalls for markets.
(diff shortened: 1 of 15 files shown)


packages/core/contracts/InsuranceFund.sol
```diff
@@ -32,15 +33,25 @@ contract InsuranceFund is IInsuranceFund, Ownable {
...
 
     /// @inheritdoc IInsuranceFund
     function resolve(IMarket market, address account) external onlyOwner isMarketInstance(market) {
         token.approve(address(market));
         market.settle(account);
-        Fixed6 resolutionAmount = market.margin().crossMarginBalances(account).mul(Fixed6Lib.NEG_ONE);
-        market.update(account, Fixed6Lib.ZERO, resolutionAmount, address(0));
+        IMargin margin = market.margin();
+        Fixed6 resolutionAmount = margin.isolatedBalances(account, market).mul(Fixed6Lib.NEG_ONE);
+        if (resolutionAmount.isZero()) { // Cross-margin shortfall
```
kbrizzle, line 48: If the account has a positive isolated balance then this will be skipped.

What if we just split this into two separate `resolve()` function, one for iso and one for cross?

packages/core/contracts/Market.sol
```diff
@@ -354,12 +354,11 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         emit RiskParameterUpdated(newRiskParameter);
     }
 
-    /// @notice Claims any available fee that the sender has accrued
-    /// @dev Applicable fees include: protocol, oracle, risk, donation, and claimable
+    /// @notice Transfers a fee that the sender has accrued to the Margin contract
+    /// @dev Applicable fees include: protocol, oracle, and risk
     /// @return feeReceived The amount of the fee claimed
```
kbrizzle, line 359: Can we put a dev note on this that it's only for post-v2.4 migration now, and can be removed once all fees have been claimed?

Also, we should add a follow up ticket to decide on a different privilege for this function so that a migrator could trigger this on behalf of the account.

EdNoepel: The only fees no longer claimed with this method are liquidation, subtractive, and solver fees which used to accumulate in a user's `Local`.  This method is still used for claiming protocol, oracle, and risk fees, which accumulate in `Global`.

packages/core/contracts/InsuranceFund.sol
EdNoepel: That was intentional.  The conditional checks whether the market is isolated for the account.  If not, (line 48) funds are moved from _InsuranceFund_ into the user's cross-margin balance, improving their collateralization across all markets.  If so (line 51), funds are moved from _InsuranceFund_ into the user's isolated balance for the market, improving their collateralization for that market.

Could you describe the behavior you're looking for?

EdNoepel: Use case of concern is if user has a cross-margin shortfall, but has a positive isolated balance in the specified market.  Furthermore, the cross-margin use case would not warrant the caller passing a _market_ parameter.

packages/core/contracts/Market.sol
EdNoepel: Per discussion earlier today, will leave this as-is.

packages/core/contracts/InsuranceFund.sol
EdNoepel: Split in af551d8b3647ab4d22085a1527ae2330bacdbf0f.

PR #556, opened 2025-01-30
SolverVault liquidation and insolvency integration tests

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-2149/[testing]-add-liquidation-and-insolvency-integration-tests-to

### DIFF
--- a/packages/vault/test/integration/vault/SolverVault.test.ts
+++ b/packages/vault/test/integration/vault/SolverVault.test.ts
@@ -1586,206 +1586,265 @@ describe('SolverVault', () => {
     })
 
     context('liquidation', () => {
-      context('short exposure', () => {
-        it('closes position under min margin to redeem liquidatable vault', async () => {
-          expect(await vault.convertToAssets(parse6decimal('1'))).to.equal(parse6decimal('1'))
-          expect(await vault.convertToShares(parse6decimal('1'))).to.equal(parse6decimal('1'))
+      it('recovers from liquidation', async () => {
+        expect(await vault.convertToAssets(parse6decimal('1'))).to.equal(parse6decimal('1'))
+        expect(await vault.convertToShares(parse6decimal('1'))).to.equal(parse6decimal('1'))
 
-          const smallDeposit = parse6decimal('10')
-          await vault.connect(user).update(user.address, smallDeposit, 0, 0)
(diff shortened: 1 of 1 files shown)


packages/vault/test/integration/vault/SolverVault.test.ts
```diff
@@ -1584,6 +1584,439 @@ describe('SolverVault', () => {
...
+          // Ensure maintenance requirement is violated.
+          const collateral = (await btcMarket.locals(vault.address)).collateral
+          const maintenanceRatio = (await btcMarket.riskParameter()).maintenance
+          const maintenanceRequired = (await btcMarket.positions(vault.address)).short
+            .mul(newBtcOraclePrice)
+            .mul(maintenanceRatio)
+            .div(1e12)
+          expect(collateral).to.be.lessThan(maintenanceRequired)
+
+          // Redeem shares to close position
```
kbrizzle, line 1676: I think we want to maintain the same flow / checks as the `MakerVault` equivalent tests here.

What was reasoning for switching this to testing redeeming a liquidatable position, instead of testing actually liquidating the position and the aftermath?

prateekdefi: Reasoning to add redeem was that I checked that liquidation will be successful by `expect(collateral).[to.be](http://to.be/).lessThan(maintenanceRequired)` and I thought of testing whether collateral from a liquidatable vault can be redeemed. Now that I think I could have added different tests for both of these two scenarios.

prateekdefi: I will update this current test to actually liquidating the position and also create a new test for redeeming the liquidatable position similar to `rebalance position to redeem liquidatable vault` in `MakerVault`.

kbrizzle: Ahh I see where you sourced these from now actually. We can actually skip the first two `short exposure` tests related to rebalancing (those are specific to the MakerVault strategy).

The two `recovers from...` tests in each group are the ones we should be referencing. Because we want to make sure if one of the positions gets liquidated, the Vault still operates all the edge cases as expected (with shortfall as well).

kbrizzle (dismissed):

EdNoepel (approved):
LGTM; will also need to merge this to v2.5.

PR #557, opened 2025-01-30
v2.5

EdNoepel:
For CI purposes.


EdNoepel:
Changed base from `main` to `v2.4-mini` by request from @kbrizzle.

packages/vault/contracts/SolverVault.sol
```diff
@@ -43,7 +49,7 @@ contract SolverVault is ISolverVault, Vault {
         to.update(address(this), Fixed6Lib.ZERO, Fixed6Lib.from(1, amount), address(0));
     }
 
-    modifier onlyCoordinator {
+    modifier onlyCoordinator override {
```
kbrizzle, line 52: nit: I don't think you need to override a modifier?

packages/vault/contracts/libs/MakerStrategyLib.sol
```diff
@@ -203,4 +210,13 @@ library MakerStrategyLib {
         marketContext.maxPosition = marketContext.currentAccountPosition.maker
             .add(marketContext.riskParameter.makerLimit.unsafeSub(marketContext.currentPosition.maker));
     }
+
+    function _getTargetPosition(MarketMakerStrategyContext memory marketContext, UFixed6 marketAssets) private pure returns (Fixed6) {
```
kbrizzle, line 214: I think this function is unused

.github/workflows/gas_report_and_code_size.yml
```diff
@@ -69,7 +69,7 @@ jobs:
             const prNumber = context.payload.pull_request.number;
 
             const isWorkflowComment = (comment) => {
-              return (comment.body.includes('### Code Size Diff:') || comment.body.includes('### Gas Report Diff:'));
+              return (comment.body.includes('### Code Size:') || comment.body.includes('### Gas Report:'));
```
prateekdefi, line 72:
```suggestion
              return (comment.body.includes('### Code Size Diff:') || comment.body.includes('### Gas Report Diff:'));
```

prateekdefi (commented):
I reviewed if the changes done by me have been merged cleanly, following is the list:
- InsuranceFund is added
- Absolute update is removed
- close method is added
- MarketFactory Protocol Parameters is updated with minMinMaintenance
- MagicValues are removed from Market
- Split maxSettlementFee into Max sync and async fee 
- MultiInvoker's trigger order is removed
- Leverage buffer parameter is added in Vault
- Account allowed settings for permissioned vaults are added
- Operator settings are consolidated to only MarketFactory
- The risk coordinator added to the vault

All of the above changes have been merged cleanly. Just added one comment regarding code size and gas diff workflow

packages/vault/contracts/SolverVault.sol
EdNoepel: I get a compilation error without marking base modifier `virtual` and overriding.  I could change the name to `onlySolverCoordinator` or such.

packages/vault/contracts/libs/MakerStrategyLib.sol
EdNoepel: removed in f13421e6c0ae54fe02f643d0fe059c5544eea34c

PR #559, opened 2025-02-05
Update: BSL-1.1 License addition

l3wi:
Update license file to cover repo files at commit height.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,96 @@
+Business Source License 1.1
+
+License text copyright (c) 2017 MariaDB Corporation Ab, All Rights Reserved.
+"Business Source License" is a trademark of MariaDB Corporation Ab.
+
+---
+
+Parameters
+
+Licensor: Perennial Labs, Inc.
+
+Licensed Work: Perennial V2 @ Commit 2d5a012c94c29953deb7da7999d326b6e850ab2d
+The Licensed Work is (c) 2024 Perennial Labs
+
+Change Date: The earlier of 2027-11-04
+
+Change License: GNU General Public License v2.0 or later
+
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -0,0 +1,96 @@
...
+---
+
+Parameters
+
+Licensor: Perennial Labs, Inc.
+
+Licensed Work: Perennial V2 @ Commit 2d5a012c94c29953deb7da7999d326b6e850ab2d
+The Licensed Work is (c) 2024 Perennial Labs
+
+Change Date: The earlier of 2027-11-04
```
EdNoepel, line 15: Was the change date supposed to be the earlier of two dates?  What is the second reference date?

LICENSE.md
```diff
@@ -0,0 +1,96 @@
...
+License text copyright (c) 2017 MariaDB Corporation Ab, All Rights Reserved.
+"Business Source License" is a trademark of MariaDB Corporation Ab.
+
+---
+
+Parameters
+
+Licensor: Perennial Labs, Inc.
+
+Licensed Work: Perennial V2 @ Commit 2d5a012c94c29953deb7da7999d326b6e850ab2d
```
EdNoepel, line 12: Do we need to state a commit?  Seems license should apply to any/all versions.

kbrizzle: Yeah I think this should implicitly cover whatever's in the repo at that time, it's already tied to commit(s) naturally through it being in the git repo.

l3wi:
@EdNoepel @kbrizzle would you be able to tag the 2.4 release commit, modify the date to +3years from the commit date & merge into the repo.

PR #560, opened 2025-02-06
Collateral Account - handle redeemPrice < 1 when unwrapping

EdNoepel:
`withdraw` now treats `amount` parameter as a maximum rather than absolute - This is cleaner than adding several calculations in the case where `shouldUnwrap == true`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,6 +7,7 @@ on:
     branches:
       - main
       - v2.4
+      - v2.4-mini
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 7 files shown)


prateekdefi (commented):
As Kevin suggested on [Slack](https://equilibriacollective.slack.com/archives/C08C55XA4PP/p1738913885451089), this issue can be fixed without the interface change. Other changes look good to me.

EdNoepel:
> this issue can be fixed without the interface change

Agree; reverted in [ce0671d](https://github.com/equilibria-xyz/perennial-v2/pull/560/commits/ce0671d4c84a035a9d1a68705e0609e948446103).

PR #562, opened 2025-02-07
Vault target array length check

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-2216/vault-does-not-validate-target-array-length


EdNoepel (commented):
To test, I would have implemented a simple strategy in `contracts/test` whose `allocate` method creates arrays of incorrect length.

PR #563, opened 2025-02-07
Prevent overwriting coordinator shares when settling

EdNoepel:
Code for awarding profit shares assumes _account_ and _coordinator_ are two separate actors.  But if account == coordinator, the `_credit` call has no effect because shares are later overwritten when context is stored.  To resolve this, pass account being settled into `_credit` and update context if receiver account is same account being settled.


kbrizzle (approved):

prateekdefi (approved):
LGTM

PR #566, opened 2025-02-09
[Audit Fix] Take intent fees into account in invariant

kbrizzle:
Approximates the outstanding intent trade fees by measuring all outstanding trade fees.

Applies these to:
- margin check
- maintenance check
- collateral below zero check

Resolves: https://audits.sherlock.xyz/contests/764/voting/31.


EdNoepel (approved):
LGTM

PR #569, opened 2025-02-11
add hardhat task to verify all contracts

prateekdefi:


arjun-io (dismissed):

packages/deploy/tasks/verifyContracts.ts
```diff
@@ -0,0 +1,25 @@
...
+    console.log('[Verify Contract] Running Verify Contract Task')
+    const {
+      deployments: { all },
+    } = HRE
+
+    const deployments = await all()
+
+    for (const deployment of Object.keys(deployments)) {
+      console.log(`Verifying: ${deployment} at ${deployments[deployment].address}`)
+      await HRE.run('verify:verify', {
```
EdNoepel, line 15: Seems there would be some value adding error handling here, such that a single failure doesn't break the loop.

EdNoepel (commented):
Not sure I understand how this works with respect to deployments.  The `Object.keys` suggests the task iterates through different chains.  But hardhat would be reading the same `ETHERSCAN_API_KEY` key from the environment.  Sholudn't the deployed chain be a task argument, to give the user a chance to set their environment appropriately for the target chain?

prateekdefi:
> Not sure I understand how this works with respect to deployments. The `Object.keys` suggests the task iterates through different chains. But hardhat would be reading the same `ETHERSCAN_API_KEY` key from the environment. Sholudn't the deployed chain be a task argument, to give the user a chance to set their environment appropriately for the target chain?

The chain is set with the `FORK_NETWORK` key from the environment. So `Object.keys` actually iterate through all contracts deployed on the selected chain. Also, the Etherscan API key is network dependent but we need a custom network for appchain network in the common hardhat config. Will add it

arjun-io (approved):

PR #570, opened 2025-02-12
[Code Size] Clean up .next() methods

kbrizzle:
Removes `.next()` methods on `Order` and `Guarantee` in favor of `.fresh()` methods which instantiate new copies of these objects instead.

This causes a very slight gas increase w/ a considerable reduction in code siz: `120 bytes`.


EdNoepel (approved):
LGTM

PR #576, opened 2025-02-17
Add synthetic orderbook

kbrizzle:
#### Migration Note
- Removes exposure
  - Need to set adiabatic fees to zero and settle exposure ahead of migration
- Changes storage layout of `Version` and `Guarantee`
  - Need to settle-only / settle all markets & vaults


EdNoepel (approved):
I pushed a commit to fix tests and make one natspec comment valid.  No other issues.

PR #577, opened 2025-02-17
Cross-margin logic

EdNoepel:
This implements collections for tracking and iterating through cross-margined markets, along with logic for cross-margin margin and maintenance checks.  This includes changes to iterate through guarantees to aggregate price adjustments, a facility which should be in place before merging v2.4 audit fixes, such that effective collateral and worst-case position may be implemented appropriately.

### DIFF
--- a/packages/core/contracts/Margin.sol
+++ b/packages/core/contracts/Margin.sol
@@ -193,7 +193,6 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
 
     /// @dev Upserts a market into cross-margin collections
     function _cross(address account, IMarket market) private {
-        // if (crossMarginMarketIndex[account][market] == 0 && (crossMarginMarkets[account].length == 0 || crossMarginMarkets[account][0] != market)) {
         if (!_isCrossed(account, market)) {
             uint256 newIndex = crossMarginMarkets[account].length;
             if (newIndex > MAX_CROSS_MARGIN_MARKETS) revert MarginTooManyCrossedMarkets();
@@ -205,9 +204,9 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
 
     /// @dev Removes a market from cross-margin collections
     function _uncross(address account, IMarket market) private {
-        uint256 index = crossMarginMarketIndex[account][market];
-        if (index != 0) {
-            uint256 lastIndex = crossMarginMarkets[account].length;
+        if (_isCrossed(account, market)) {
+            uint256 index = crossMarginMarketIndex[account][market];
+            uint256 lastIndex = crossMarginMarkets[account].length - 1;
             if (index != lastIndex) {
                 // Swap last item with the one being removed
                 IMarket lastMarket = crossMarginMarkets[account][lastIndex];
@@ -228,7 +227,6 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
         bool updateCheckpoint_
     ) private {
         // Calculate new balances
-        // FIXME: probably shouldn't revert if closing due to liquidation
         Fixed6 newCrossBalance = _balances[account][CROSS_MARGIN].sub(amount);
         if (newCrossBalance.lt(Fixed6Lib.ZERO)) revert MarginInsufficientCrossedBalance();
         Fixed6 oldIsolatedBalance = _balances[account][market];
@@ -241,8 +239,7 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
         // TODO: We could add logic here to support switching modes with a position.
         if ((isolating || crossing) && _hasPosition(account, market)) revert MarginHasPosition();
 
-        // Ensure user has not exceeded max number of crossed markets
-        if (crossing) _cross(account, market);
+        // If switching modes, remove from cross-margin collections
         if (isolating) _uncross(account, market);
 
         // update storage
--- a/packages/core/contracts/Margin.sol
+++ b/packages/core/contracts/Margin.sol
@@ -8,6 +8,8 @@ import { UFixed6, UFixed6Lib } from "@equilibria/root/number/types/UFixed6.sol";
 import { Token18, UFixed18, UFixed18Lib } from "@equilibria/root/token/types/Token18.sol";
 
 import { Checkpoint, CheckpointStorage } from "./types/Checkpoint.sol";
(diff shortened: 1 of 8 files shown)


packages/core/contracts/Margin.sol
```diff
@@ -16,13 +18,20 @@ import { IMarket, IMarketFactory } from "./interfaces/IMarketFactory.sol";
...
+    uint256 public constant MAX_CROSS_MARGIN_MARKETS = 8;
+
+    /// inheritdoc IMargin
     Token18 public immutable DSU; // solhint-disable-line var-name-mixedcase
 
     /// @dev Contract used to validate markets
     IMarketFactory public marketFactory;
 
-    // TODO: Introduce an iterable collection of cross-margained markets for a user.
+    // Iterable collection of cross-margined markets for a user (account => markets)
```
kbrizzle, line 30: nit: `/// @dev` for both of these comments

packages/core/contracts/Margin.sol
```diff
@@ -115,7 +125,12 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
 
     /// @inheritdoc IMargin
     function handleMarketUpdate(address account, Fixed6 collateralDelta) external onlyMarket {
-        _isolate(account, IMarket(msg.sender), collateralDelta, false);
+        if (!collateralDelta.isZero() || _isIsolated(account, IMarket(msg.sender)))
+            // Account's isolated collateral is changing, or market is already isolated
+            _isolate(account, IMarket(msg.sender), collateralDelta, false);
```
kbrizzle, line 130: what case here would cause `collateralDelta > 0`, but isIsolated to be false?

packages/core/contracts/Margin.sol
```diff
@@ -142,6 +157,16 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
         return _balances[account][market];
     }
 
+    /// @inheritdoc IMargin
+    function isCrossed(address account, IMarket market) external view returns (bool) {
+        return _isCrossed(account, market);
+    }
+
+    /// @inheritdoc IMargin
+    function isIsolated(address account, IMarket market) external view returns (bool) {
```
kbrizzle, line 166: can we just have one getter here: `isolated(bool)` which returns false if cross margined? or is there a case where both are false?

packages/core/contracts/Margin.sol
```diff
@@ -198,18 +288,23 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
...
+        }
+
+        // If decreasing an existing isolated balance with position, ensure price is not stale
+        if (market.hasPosition(account) && decreasingIsolatedBalance && market.stale()) {
+            revert IMarket.MarketStalePriceError();
         }
-        // TODO: if amount.sign() = 1, ensure remaining cross-margin balance is sufficient margin for all markets
 
         if (isolating) emit MarketIsolated(account, market);
-        if (crossing) emit MarketCrossed(account, market);
```
kbrizzle, line 212: is this meant to be deleted?

packages/core/contracts/Margin.sol
kbrizzle: I think the answer is we need a way to check if a market is registered in the cross margin pool vs just idle

packages/core/test/integration/core/crossMargin.test.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    // check margin
+    const marginA = await marketA.market.marginRequired(userA.address, 0)
+    expect(marginA).to.equal(parse6decimal('30300'))
+    const marginB = await marketB.market.marginRequired(userA.address, 0)
+    expect(marginB).to.equal(parse6decimal('44910'))
+
+    // cannot remove more collateral than is needed to maintain margin requirements
+    const maxWithdrawl = parse6decimal('200000').sub(marginA).sub(marginB)
+    await expect(margin.connect(userA).withdraw(userA.address, maxWithdrawl.add(1))).to.be.revertedWithCustomError(
```
kbrizzle, line 266: would be good to test the counter case on these, that you can isolate / withdraw when the amount is just below the limit.

kbrizzle:
General note on test: think we need more thorough coverage in either the `Market.test.ts` or integration tests layer, especially around having both isolated and cross positions open as well as around crossing / isolating positions and ensuring there's no weirdness with registration or collateral / fee / shortfall movements.

As with these other comments, this is all can be done in followup cleanups as well if that makes more sense.

packages/core/contracts/Margin.sol
EdNoepel: Yes, the market is not crossed until added to the cross-market collections.

packages/core/contracts/Margin.sol
EdNoepel: Use case: the market is not currently isolated, and this is the first update which occurs with user specifies collateral in the `Order`.

packages/core/contracts/Margin.sol
EdNoepel: fixed in 58e3748baaa5e77ec409be52effa185240c7144d

packages/core/test/integration/core/crossMargin.test.ts
EdNoepel: Added in 58e3748baaa5e77ec409be52effa185240c7144d

EdNoepel:
> think we need more thorough coverage in either the `Market.test.ts` or integration tests layer

Definitely agree; it's not ready.  Only did this PR to reduce merge complexity from v2.4 audit fixes.

packages/core/contracts/Margin.sol
kbrizzle: What would happen here if the market was in cross mode and this was called with a non-zero collateral?

I'm guessing an invariant somewhere else would revert in that case?

kbrizzle (approved):

packages/core/contracts/Margin.sol
EdNoepel: Depends; if there is a position or isolating would break margin requirements (either for this market, or for the other crossed markets, which would have less collateral available), it would revert.  If there is no position, and margin requirements are met, it should transition from crossed to isolated mode.

PR #579, opened 2025-02-19
Merge audit fixes

EdNoepel:
Merge audit fixes from `main` into `v2.5`.

These fixes included changes to margin and maintenance checks.  Originally implemented in `InvariantLib`, logic from these changes has been moved into _Market_ and applied to `marginRequired` and `maintenanceRequired` calls made by `Margin` contract for both isolated and cross-margin markets.

Upon resolving an issue with the merge of zero-cross tests, an issue was observed with `VersionLib._accumulateSpread`.  This issue is documented in the tests and affected tests have been skipped.


packages/core/contracts/Market.sol
```diff
@@ -635,26 +667,23 @@ contract Market is IMarket, Instance, ReentrancyGuard {
 
         // when liquidating, ensure maintenance requirements are not met
         if (newOrder.protected() && (
-            !context.pendingLocal.neg().eq(context.latestPositionLocal.magnitude()) || // total pending close is not equal to latest position
+            context.pendingLocal.neg().lt(context.latestPositionLocal.magnitude()) ||  // total pending close is not equal to latest position
             margin.maintained(context.account) ||                                      // latest position is properly maintained
-            !newOrder.collateral.eq(Fixed6Lib.ZERO)                                    // TODO: can eliminate because close method doesn't touch collateral
-        )) revert IMarket.MarketInvalidProtectionError();
+            !newOrder.collateral.eq(Fixed6Lib.ZERO) ||                                 // TODO: can eliminate because close method doesn't touch collateral
+            !newOrder.pos().eq(UFixed6Lib.ZERO))                                       // TODO: can eliminate because close orders cannot increase position
```
kbrizzle, line 673: I don't think we can eliminate this one, what was the reasoning here?

packages/core/contracts/libs/InvariantLib.sol
```diff
@@ -89,21 +90,4 @@ library InvariantLib {
             newOrder.decreasesLiquidity(updateContext.currentPositionGlobal)
         ) revert IMarket.MarketInsufficientLiquidityError();
     }
-
-    function validateProtection(IMarket.Context memory context, bool maintained, Order memory newOrder) external pure returns (bool) {
```
kbrizzle, line 93: did this all get ported into `Market.sol`? I don't think I see crossing-zero logic in the new code, or did you remove that on purpose to get things working?

packages/core/contracts/Market.sol
EdNoepel: Regarding the `TODO` comments: Protected orders can only be created from the `close` method, which has no parameters for changing collateral or position.  As such, `newOrder.collateral` and `newOrder.pos` should always be 0 when `newOrder.protected()` is true.

kbrizzle: Oh good call, forgot that we transitioned this to the close method.

I wonder if we should still keep this in the invariant tho (i.e. for if it got re-added in the future somehow).

packages/core/contracts/libs/InvariantLib.sol
kbrizzle: Yes to the last part, we utilize no-op closes for liquidation when the position is not yet locked, but you can't actually close larger than a zero position. This is used quite a bit in the crossing zero flows.

EdNoepel: Alright; I will port this logic over.  Was there an existing test which should have failed without it?

EdNoepel: When I merged the tests, absolute `update` calls which had _protected_ set true were turned into `close` calls, regardless of whether a position change was specified.  I assume the "no-op" logic is built into the `close` method?

kbrizzle: Yeah these tests should cover this:
- https://github.com/equilibria-xyz/perennial-v2/blob/main/packages/core/test/unit/market/Market.test.ts#L24375
- https://github.com/equilibria-xyz/perennial-v2/blob/main/packages/core/test/unit/market/Market.test.ts#L24475
- https://github.com/equilibria-xyz/perennial-v2/blob/main/packages/core/test/unit/market/Market.test.ts#L24574

`close()` was added in v2.5 while the heavier use of no-op was added in v2.4-mini so there some merge work to do there. no-op liquidations were still possible in previous version, but we probably don't have test for them since they weren't special cased.

packages/core/contracts/Market.sol
```diff
@@ -667,13 +667,13 @@ contract Market is IMarket, Instance, ReentrancyGuard {
 
         // when liquidating, ensure maintenance requirements are not met
         if (newOrder.protected() && (
+            (context.pendingLocal.crossesZero() && !newOrder.isEmpty()) ||             // pending zero-cross, liquidate (lock) with no-op order
```
kbrizzle, line 670: this logic looks a bit different than `v2.4-mini`:

- if zero cross -> revert if not empty
- otherwise -> revert if not full close (specifically this clause)

packages/core/contracts/libs/InvariantLib.sol
EdNoepel: Those tests exist in this branch (`Market.test.cs` lines 22335, 22405, and 22486).  The first two run cleanly, without (before) and with (now) the updated logic.  The last test is skipped due to the underflow issue with zero-cross.

packages/core/contracts/Market.sol
EdNoepel: I see; the comparison between pending order and position should only be made if order is not crossing zero.  Will revise.

EdNoepel: Done in 205320434e3e55f938602be5c817c4cb215b2190.

kbrizzle (approved):

prateekdefi (approved):

PR #580, opened 2025-02-19
Disable retries in periphery integration test CI

EdNoepel:
When mocha tests are run with retries enabled, errors raised in `afterEach` clauses are suppressed.  This leads to misleading error messages.  In particular, when `periphery` integration tests fail due to gas configuration, CI omits the gas out-of-range error, and shows only subsequent errors for nonce/orderId mismatch.

Since retries are not useful for periphery integration tests, disable them in CI.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,6 +40,7 @@ jobs:
 
   # [CORE]
   core-unit-test:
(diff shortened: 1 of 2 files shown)


prateekdefi (approved):
LGTM! Thanks for fixing this.

PR #581, opened 2025-02-21
Additive fees

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1869/[feature]-enshrined-additive-fees

### DIFF
--- a/packages/core/contracts/Market.sol
+++ b/packages/core/contracts/Market.sol
@@ -161,6 +161,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
             address(0),
             UFixed6Lib.ZERO,
             UFixed6Lib.ZERO,
+            UFixed6Lib.ZERO,
             false
         ); // account
         _updateIntent(
@@ -171,6 +172,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
             intent.originator,
             intent.solver,
             intent.fee,
+            intent.additiveFee,
             intent.collateralization,
             true
         ); // signer
@@ -199,6 +201,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
             address(0),
             UFixed6Lib.ZERO,
             UFixed6Lib.ZERO,
+            UFixed6Lib.ZERO,
             false
         ); // solver
         _updateIntent(
@@ -209,6 +212,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
             fill.intent.originator,
             fill.intent.solver,
             fill.intent.fee,
+            fill.intent.additiveFee,
             fill.intent.collateralization,
             true
         ); // trader
@@ -219,7 +223,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     /// @param signature Signature of taker or authorized signer
     function update(Take calldata take, bytes memory signature) external nonReentrant whenNotPaused {
         verifier.verifyTake(take, signature);
-        _updateMarket(take.common.account, take.common.signer, Fixed6Lib.ZERO, take.amount, Fixed6Lib.ZERO, take.referrer);
+        _updateMarket(take.common.account, take.common.signer, Fixed6Lib.ZERO, take.amount, Fixed6Lib.ZERO, take.referrer, UFixed6Lib.ZERO);
     }
 
     /// @notice Updates the account's position
@@ -228,7 +232,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
     /// @param takerAmount The position delta of the order (positive for long, negative for short)
     /// @param referrer The referrer of the order
     function update(address account, Fixed6 takerAmount, address referrer) external nonReentrant whenNotPaused {
-       _updateMarket(account, msg.sender, Fixed6Lib.ZERO, takerAmount, Fixed6Lib.ZERO, referrer);
+       _updateMarket(account, msg.sender, Fixed6Lib.ZERO, takerAmount, Fixed6Lib.ZERO, referrer, UFixed6Lib.ZERO);
     }
 
     /// @notice Updates the account's taker position and isolated collateral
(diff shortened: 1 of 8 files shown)


packages/core/contracts/types/Order.sol
```diff
@@ -497,6 +509,7 @@ library OrderStorageGlobalLib {
 ///         uint64 makerReferral;
 ///         uint1 protection;
 ///         uint8 invalidation;
+///         uint64 additiveFee;
```
kbrizzle, line 512: Since we have to do the whole settle-up migration anyways, maybe we put this right after `makerReferral`?

packages/core/contracts/types/Order.sol
```diff
@@ -526,11 +524,11 @@ library OrderStorageLocalLib {
             UFixed6.wrap(uint256(slot1 << (256 - 64 - 64)) >> (256 - 64)),
             UFixed6.wrap(uint256(slot1 << (256 - 64 - 64 - 64)) >> (256 - 64)),
             UFixed6.wrap(uint256(slot1 << (256 - 64 - 64 - 64 - 64)) >> (256 - 64)),
-            uint256(slot2 << (256 - 64 - 64 - 1)) >> (256 - 1),
-            uint256(slot2 << (256 - 64 - 64 - 1 - 8)) >> (256 - 8),
+            uint256(slot2 << (256 - 64 - 64 - 64 - 1)) >> (256 - 1),
+            uint256(slot2 << (256 - 64 - 64 - 64 -1 - 8)) >> (256 - 8),
```
kbrizzle, line 528: nit: space

packages/periphery/contracts/TriggerOrders/types/TriggerOrder.sol
```diff
@@ -22,6 +22,8 @@ struct TriggerOrder {
     bool isSpent;
     /// @dev Passed to market for awarding referral fee
     address referrer;
+    /// @dev Passed to market for awarding additive fee to referrer
+    UFixed6 additiveFee;
     /// @dev Additive fee optionally awarded to GUIs upon execution
     InterfaceFee interfaceFee;
```
kbrizzle, line 28: This change actually allows us to remove the interface fee, let's clean that up as well here.

(The `InterfaceFee` was our old hacky way of doing an additive fee with trigger orders).

packages/periphery/contracts/TriggerOrders/types/TriggerOrder.sol
```diff
@@ -134,7 +136,8 @@ struct StoredTriggerOrder {
     bytes5 __unallocated0__;      // padding for 32-byte alignment
     /* slot 1 */
     address referrer;
-    bytes12 __unallocated1__;     // padding for 32-byte alignment
+    uint64 additiveFee;           // < 18.45t
```
kbrizzle, line 139: Since this is just storing the % fee unlike the field in `Order` we only need uint24 here.

packages/core/contracts/Market.sol
```diff
@@ -240,7 +244,18 @@ contract Market is IMarket, Instance, ReentrancyGuard {
...
+
+    function update(
+        address account,
+        Fixed6 makerAmount,
+        Fixed6 takerAmount,
+        Fixed6 collateral,
+        address referrer,
+        UFixed6 additiveFee
+    ) external nonReentrant whenNotPaused {
+        _updateMarket(account, address(0), makerAmount, takerAmount, collateral, referrer, additiveFee);
```
kbrizzle, line 258: We actually need to validate that `additiveFee < ProtocolParameter.maxFee` here like we do the solverFee in intents: https://github.com/equilibria-xyz/perennial-v2/blob/main/packages/core/contracts/Market.sol#L133.

packages/periphery/contracts/TriggerOrders/types/TriggerOrder.sol
prateekdefi: I have not removed the interface fee here as we allow the user to pass the referrer and interface fee receiver separately. If we remove the interface fee, we would end up sending both the referrer and the additive fee to the same referrer unlike previously. I'm not sure if we actually want this.

prateekdefi: I have not removed the interface fee here as we allow the user to pass the referrer and interface fee receiver separately. If we remove the interface fee, we would end up sending both the referrer and the additive fee to the same referrer unlike previously. I'm not sure if we actually want this. @kbrizzle

kbrizzle: Yep, we no longer really need two separate referrer addresses either, they both should go to the same place.

packages/core/contracts/Market.sol
prateekdefi: Added in [c2b82af](https://github.com/equilibria-xyz/perennial-v2/pull/581/commits/c2b82af00a428e22beaf953ac9d618b163f7d4aa)

packages/periphery/contracts/TriggerOrders/types/TriggerOrder.sol
prateekdefi: Updated in [c2b82af](https://github.com/equilibria-xyz/perennial-v2/pull/581/commits/c2b82af00a428e22beaf953ac9d618b163f7d4aa)

packages/periphery/contracts/TriggerOrders/types/TriggerOrder.sol
prateekdefi: Removed interface fee in [c2b82af](https://github.com/equilibria-xyz/perennial-v2/pull/581/commits/c2b82af00a428e22beaf953ac9d618b163f7d4aa)

packages/periphery/contracts/TriggerOrders/types/TriggerOrder.sol
```diff
@@ -136,14 +134,8 @@ struct StoredTriggerOrder {
...
-    uint64 additiveFee;           // < 18.45t
-    bytes4 __unallocated1__;     // padding for 32-byte alignment
-    /* slot 2 */
-    address interfaceFeeReceiver;
-    uint64 interfaceFeeAmount;    // < 18.45t
-    bool interfaceFeeFlat;
-    bool interfaceFeeUnwrap;
-    // 2 bytes left over (no need to pad trailing bytes)
+    uint24 additiveFee;           // <= 1677%
+    bytes9 __unallocated1__;     // 9 bytes left over (no need to pad trailing bytes)
```
kbrizzle, line 138: nit: let's still leave a comment here that slot 2 can be almost fully occupied by dirty data until it is re-stored.

kbrizzle (approved):

PR #587, opened 2025-03-03
Auto-deisolate on position close

EdNoepel:
Changes for PE-1561.  When positions close, markets are implicitly deisolated.  Added a switch to disable this behavior, currently used for isolated Vaults.  Can revisit that when cross-margin checkpoints are implemented.

### DIFF
--- a/packages/core/contracts/Margin.sol
+++ b/packages/core/contracts/Margin.sol
@@ -7,6 +7,7 @@ import { Fixed6, Fixed6Lib } from "@equilibria/root/number/types/Fixed6.sol";
 import { UFixed6, UFixed6Lib } from "@equilibria/root/number/types/UFixed6.sol";
 import { Token18, UFixed18, UFixed18Lib } from "@equilibria/root/token/types/Token18.sol";
 
+import { CheckpointLib } from "./libs/CheckpointLib.sol";
 import { Checkpoint, CheckpointStorage } from "./types/Checkpoint.sol";
 import { Guarantee } from "./types/Guarantee.sol";
 import { Local } from "./types/Local.sol";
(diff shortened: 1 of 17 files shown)


packages/core/contracts/Margin.sol
```diff
@@ -45,6 +45,12 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
     /// Cross-margin checkpoints stored as IMarket(address(0))
     mapping(address => mapping(IMarket => mapping(uint256 => CheckpointStorage))) private _checkpoints;
 
+    /// @notice Supresses default behavior of deisolating funds when a position is closed
+    mapping(address => bool) public autoDeisolateDisabled;
```
kbrizzle, line 49: nit: maybe add a TODO that this ideally should be remove if we can resolve the Vault issues

packages/vault/contracts/libs/MakerStrategyLib.sol
```diff
@@ -87,19 +87,17 @@ library MakerStrategyLib {
 
     /// @notice Compute the target allocation for each market
     /// @param registrations The registrations of the underlying markets
-    /// @param deposit The amount of assets that are being deposited into the vault
     /// @param withdrawal The amount of assets to make available for withdrawal
     /// @param ineligible The amount of assets that are inapplicable for allocation
     function allocate(
         Registration[] memory registrations,
-        UFixed6 deposit,
```
kbrizzle, line 95: why is deposit being removed? shouldn't we still have that at the Vault level, even if at the market level we're only isolating / deisolating?

packages/core/contracts/Margin.sol
```diff
@@ -124,16 +135,37 @@ contract Margin is IMargin, Instance, ReentrancyGuard {
...
-            _cross(account, IMarket(msg.sender));
+            _cross(account, market);
     }
 
     /// @inheritdoc IMargin
-    function updateClaimable(address account, Fixed6 collateralDelta) external onlyMarket {
+    function handleMarketSettle(address account, uint256 latestVersion) external onlyMarket {
+        IMarket market = IMarket(msg.sender);
+        UFixed6 isolatedBalance = UFixed6Lib.unsafeFrom(_balances[account][market]);
+        // If market also had no position when isolated balance last changed, should not deisolate
```
kbrizzle, line 151: Is this so that we only "try sweeping" balances once per the conditions set?

packages/vault/contracts/libs/MakerStrategyLib.sol
kbrizzle: Changing any of the Vault accounting logic is pretty scary so if we need this, we may need to do some extra thinking here.

packages/core/contracts/Margin.sol
EdNoepel: done in f7856ff32e06329fdcf4a391eb8427d084208281

packages/vault/contracts/libs/MakerStrategyLib.sol
EdNoepel: In my almost-working implementation, it made more sense to read from cross-margin balance.  Reverted all these changes in f7856ff32e06329fdcf4a391eb8427d084208281, and instead disabled close-on-deisolate for both _MakerVault_ and _SolverVault_.

packages/core/contracts/Margin.sol
EdNoepel: The comment on 151 addresses the use case where `isolate` is called with no position and market is settled with no position afterward.  Expected behavior is that user's funds should remain isolated; the user didn't _close a position_, they simply _have no position_.

packages/vault/contracts/libs/MakerStrategyLib.sol
kbrizzle: I see, yeah let me do this as part of the cross margin vault implementation so that all those changes are in one place 👍

kbrizzle (approved):

PR #593, opened 2025-03-06
prevent _processOrderLocal call for 0 address

EdNoepel:
- Global settlement invoked by oracle after a price commitment is making lots of seemingly unnecessary `_processOrderLocal` calls.  Found it in my branch, where `Margin` contract was getting calls to write checkpoints for the 0 address.  To optimize for space instead of gas, we could instead have `_processOrderLocal` immediately return if account is 0 address.
- Sidequest: Fixed `multi-delayed update w/ collateral (gas)` integration test, which was not functional.
- Resolved some unrelated linter warnings.

### DIFF
--- a/packages/core/contracts/Market.sol
+++ b/packages/core/contracts/Market.sol
@@ -868,7 +868,6 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         ) _processOrderGlobal(context, settlementContext, context.global.latestId + 1, nextOrder.timestamp, nextOrder);
 
         while (
-            context.account != address(0) &&
             context.local.currentId != context.local.latestId &&
             (nextOrder = _pendingOrders[context.account][context.local.latestId + 1].read()).ready(context.latestOracleVersion)
         ) _processOrderLocal(context, settlementContext, context.local.latestId + 1, nextOrder.timestamp, nextOrder);
--- a/packages/core/contracts/Market.sol
+++ b/packages/core/contracts/Market.sol
@@ -868,6 +868,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         ) _processOrderGlobal(context, settlementContext, context.global.latestId + 1, nextOrder.timestamp, nextOrder);
 
         while (
+            context.account != address(0) &&
             context.local.currentId != context.local.latestId &&
             (nextOrder = _pendingOrders[context.account][context.local.latestId + 1].read()).ready(context.latestOracleVersion)
         ) _processOrderLocal(context, settlementContext, context.local.latestId + 1, nextOrder.timestamp, nextOrder);
@@ -886,7 +887,7 @@ contract Market is IMarket, Instance, ReentrancyGuard {
                 _pendingOrder[context.global.latestId].read()
             );
 
-        if (context.latestOracleVersion.timestamp > context.latestPositionLocal.timestamp)
+        if (context.account != address(0) && context.latestOracleVersion.timestamp > context.latestPositionLocal.timestamp)
             _processOrderLocal(
                 context,
                 settlementContext,
--- a/packages/core/test/integration/core/closedProduct.test.ts
+++ b/packages/core/test/integration/core/closedProduct.test.ts
@@ -122,7 +122,6 @@ describe('Closed Market', () => {
     await chainlink.nextWithPriceModification(price => price.mul(4))
     await chainlink.nextWithPriceModification(price => price.mul(4))
 
-    const LIQUIDATION_FEE = BigNumber.from('1000000000')
     await market.connect(user).close(user.address, true, constants.AddressZero)
     await market.connect(userB).close(userB.address, true, constants.AddressZero)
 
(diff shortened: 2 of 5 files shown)


kbrizzle:
Yeah interesting catch. So we do have [something similar](https://github.com/equilibria-xyz/perennial-v2/blob/main/packages/vault/contracts/Vault.sol#L401) in `Vault` as well.

I think the settle part of the flow cannot be hit (we don't ever have `localId / currentId > 0` on the address(0) account), but it does appear that the sync portion of the flow would continually get hit.

I'm curious why this didn't seem to affect the gas diff though?

EdNoepel:
> I'm curious why this didn't seem to affect the gas diff though?

I assume this is because the oracle setup in most `core` integration tests is a mocked chainlink implementation which does not invoke global settlement upon price commitment.

kbrizzle:
I'm surprised we don't ever call `.settle(address(0))` in the integration tests 🤔 

Could you maybe a test case to this would trigger it? Just want to verify it's actually cut out gas in some situation.

EdNoepel:
> Could you maybe add a test case to this would trigger it? Just want to verify it's actually cut out gas in some situation.

Performed two experiments highlighting the gas savings.

First, after fixing `delayed update w/ collateral (gas)` test, I added a `settle(constants.AddressZero)` inside the loop and ran with and without the 0-check.
'delayed update w/ collateral (gas)', settle(address(0)) in loop, without 0-check:

```
································|····················|······················|····················|·····························
|  Solidity: 0.8.24             ·  Optim: true       ·  Runs: 1000000       ·  viaIR: true       ·   Block: 32,000,000 gas    │
································|····················|······················|····················|·····························
|  Methods                                                                                                                    │
································|····················|······················|····················|··············|··············
|  Contracts / Methods          ·  Min               ·  Max                 ·  Avg               ·  # calls     ·  usd (avg)  │
································|····················|······················|····················|··············|··············
|  Market                       ·                                                                                             │
································|····················|······················|····················|··············|··············
|      settle                   ·           191,370  ·             554,270  ·           412,455  ·          10  ·          -  │
································|····················|······················|····················|··············|··············
```

'delayed update w/ collateral (gas)', settle(address(0)) in loop, with 0-check in _processOrderLocal

```
································|··················|···················|····················|·····························
|  Solidity: 0.8.24             ·  Optim: true     ·  Runs: 1000000    ·  viaIR: true       ·   Block: 32,000,000 gas    │
································|··················|···················|····················|·····························
|  Methods                                                                                                               │
································|··················|···················|····················|··············|··············
|  Contracts / Methods          ·  Min             ·  Max              ·  Avg               ·  # calls     ·  usd (avg)  │
································|··················|···················|····················|··············|··············
|  Market                       ·                                                                                        │
································|··················|···················|····················|··············|··············
|      settle                   ·  198,943 +7,573  ·  471,325 -82,945  ·   354,368 -58,087  ·          10  ·          -  │
································|··················|···················|····················|··············|··············
```

Second, I modified _setupHelpers_ `settle` method to `settle(constants.AddressZero)`, just as the oracle would, and run `Happy Path` tests:

Happy Path, settle(0) in settle helper, without 0-check in _settle
```
·······································|····················|························|····················|·····························
|  Solidity: 0.8.24                    ·  Optim: true       ·  Runs: 1000000         ·  viaIR: true       ·   Block: 32,000,000 gas    │
·······································|····················|························|····················|·····························
|  Market                              ·                                                                                               │
·······································|····················|························|····················|··············|··············
|      settle                          ·           183,995  ·               705,596  ·           407,420  ·          44  ·          -  │
·······································|····················|························|····················|··············|··············
```

Happy Path, settle(0) in settle helper, with 0-check in _settle
```
·······································|··················|····················|····················|·····························
|  Solidity: 0.8.24                    ·  Optim: true     ·  Runs: 1000000     ·  viaIR: true       ·   Block: 32,000,000 gas    │
·······································|··················|····················|····················|··············|··············
|  Market                              ·                                                                                         │
·······································|··················|····················|····················|··············|··············
|      settle                          ·     184,041 +46  ·  590,780 -114,816  ·   362,251 -45,169  ·          44  ·          -  │
·······································|··················|····················|····················|··············|··············
```

kbrizzle (approved):

EdNoepel:
`Liquidation` integration tests run fine on their own, but fail when run with other integration tests.  I can reproduce in my working copy and am currently investigating.

EdNoepel:
> I think the settle part of the flow cannot be hit (we don't ever have `localId / currentId > 0` on the address(0) account)
I confirmed this and reverted the change.

EdNoepel:
> `Liquidation` integration tests run fine on their own, but fail when run with other integration tests. I can reproduce in my working copy and am currently investigating.

I re-skipped the `multi-delayed update w/ collateral (gas)` test to work around this.  I suspect something in the oracle fakes is not resetting in `ChainlinkContext.reset()`, but I burned too much time trying to fix it.

PR #601, opened 2025-03-17
Remove collateralization field from intent

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-1899/[cleanup]-remove-collateralization-field-from-intent

### DIFF
--- a/packages/core/contracts/Market.sol
+++ b/packages/core/contracts/Market.sol
@@ -148,7 +148,6 @@ contract Market is IMarket, Instance, ReentrancyGuard {
                 guaranteeReferrer: address(0),
                 guaranteeReferralFee: UFixed6Lib.ZERO,
                 additiveFee: UFixed6Lib.ZERO,
-                collateralization: UFixed6Lib.ZERO,
                 chargeTradeFee: false
             })
         ); // account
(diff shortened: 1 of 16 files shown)


kbrizzle (approved):

packages/core/contracts/types/Order.sol
```diff
@@ -97,6 +97,7 @@ library OrderLib {
     /// @param collateral The change in the collateral
     /// @param protect True when liquidating the position
     /// @param referralFee The referral fee
+    /// @param additiveFee The additive fee for the order
```
EdNoepel, line 100: Self-referential comment; a better comment would answer the question "what is an additive fee"?

packages/core/contracts/Market.sol
```diff
@@ -606,7 +599,6 @@ contract Market is IMarket, Instance, ReentrancyGuard {
         updateContext.liquidator = liquidators[context.account][context.local.currentId];
         updateContext.orderReferrer = orderReferrers[context.account][context.local.currentId];
         updateContext.guaranteeReferrer = guaranteeReferrers[context.account][context.local.currentId];
-        updateContext.collateralization = collateralization;
```
EdNoepel, line 609: Can we remove `collateralization` from `UpdateContext` entirely?

EdNoepel (commented):
Comment inline regarding `UpdateContext`.  Also probably needs to be removed from `IMargin.margined`, but may be best to pass zeros instead since that implementation is being heavily refactored.

packages/core/contracts/Market.sol
prateekdefi: Removed in [b0953f1](https://github.com/equilibria-xyz/perennial-v2/pull/601/commits/b0953f1c89485e8d3506db56d62816709fd1c895). Was unsure if it might be useful in implementing cross margin.

PR #602, opened 2025-03-17
Additional "Happy Path" integration tests

EdNoepel:
- There was already a test to cover invalid price, but I extended it to ensure a new position settled after commiting a valid price.
- Added a new short position test which checks state on both opening and closing the position.
- Created a test which fakes a stale oracle price, ensuring opening and closing a maker order is rejected when price is stale.


kbrizzle (approved):

prateekdefi (approved):
LGTM

PR #605, opened 2025-03-20
Remove unused networks and deployments

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-2359/[repo]-remove-unused-networks-and-deployments

- Since we don't have a `chainLinkFeedRegistry` contract on perennial, updated chainlink helpers to create Oracle versions directly. 
- Each `next` now creates the next version of the previous timestamp + 100 secs (configurable, should be decided)
- Many tests currently fail since we assert timestamps and fees based on prices from Chainlink Oracle.


packages/core/external/deployments/perennial/DSU.json
EdNoepel, line 1: We already have external deployments for DSU and USDC in [deploy/external/deployments/perennial](https://github.com/equilibria-xyz/perennial-v2/blob/main/packages/deploy/external/deployments/perennial/DSU.json); no need to redefine this.  Just update `externalDeployments` in `hardhat.config.ts`.

EdNoepel (commented):
Given new direction for v3, recommend we limit this task to a minor cleanup.  Tests which rely on chain-specific oracle data should be left alone until we're ready to rewrite them in solidity, in a more chain-agnostic manner.

prateekdefi:
> Given new direction for v3, recommend we limit this task to a minor cleanup. Tests which rely on chain-specific oracle data should be left alone until we're ready to rewrite them in solidity, in a more chain-agnostic manner.

Agreed, there are many integration tests that use oracle data, and fixing them would require significant effort. Since we are moving to Solidity in v3, we can leave them as they are for now.

Closing this PR. I’ll create a new PR for cleaning up unused network deployments

PR #606, opened 2025-03-20
Update packages for npm publish

arjun-io:
Updates the package.json versions to v1.4 to we can publish the packages.

Also adds some missing getters to IAccount - ideally the comments in Account.sol would also be updated to use inheritdoc but I didn't want to touch the impl contracts. Maybe we can update those in the next release

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "perennialmono",
-  "version": "1.3.0",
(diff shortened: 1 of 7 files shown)


EdNoepel (approved):
LGTM.  Agree with those IAccount interface changes.

kbrizzle (approved):

PR #607, opened 2025-03-25
Add some pending order checks to Market checks

EdNoepel:
This is quite incomplete, but figured there may be value merging yesterday's work before abandoning the task for more important endeavors.

### DIFF
--- a/packages/core/test/unit/market/Market.test.ts
+++ b/packages/core/test/unit/market/Market.test.ts
@@ -911,6 +911,19 @@ describe('Market', () => {
             constants.AddressZero,
           )
 
+        expectOrderEq(await market.pending(), {
+          ...DEFAULT_ORDER,
+          orders: 1,
+          collateral: COLLATERAL,
+          makerPos: POSITION,
+        })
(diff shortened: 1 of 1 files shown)


packages/core/test/unit/market/Market.test.ts
```diff
@@ -1141,7 +1179,7 @@ describe('Market', () => {
       })
     })
 
-    describe('#update', async () => {
+    describe.only('#update', async () => {
```
prateekdefi, line 1182:
```suggestion
    describe('#update', async () => {
```

prateekdefi (commented):
Just a suggestion removing `only`. Otherwise, looks good

packages/core/test/unit/market/Market.test.ts
EdNoepel: Thanks for the catch; fixed in a66725f1e74dda6442a7c52d2f9e96a9e205c93b.

kbrizzle (approved):

prateekdefi (approved):
LGTM

PR #612, opened 2025-05-09
Add FeeCoordinator / FeeSplitter

kbrizzle:


packages/periphery/contracts/Splitter/FeeCoordinator.sol
```diff
@@ -0,0 +1,65 @@
...
+        return _markets.values();
+    }
+
+    /// @notice Returns the set of created splitters.
+    function splitters() external view returns (address[] memory) {
+        return _splitters.values();
+    }
+
+    /// @notice Registers `market` as a registered market if it is valid.
+    function register(IMarket market) external {
```
EdNoepel, line 56: Guess we don't need an `onlyOwner` modifier here because we validate the market on line 57.

packages/periphery/contracts/Splitter/FeeSplitter.sol
```diff
@@ -0,0 +1,96 @@
...
+        if (totalSplit.gt(UFixed6Lib.ONE)) revert FeeSplitterOverflowError();
+
+        emit SplitUpdated(beneficiary_, newSplit);
+    }
+
+    /// @notice Claims fees from all registered markets and distributes them to the beneficiaries.
+    /// @dev Unwraps DSU into USDC before distributing.
+    function poke() external {
+        address[] memory markets = IFeeCoordinator(address(factory())).markets();
+        for (uint256 i; i < markets.length; i++) {
```
EdNoepel, line 85: Curious how many markets we could iterate through before hitting the block gas limit.  I assume this could handle one or two orders of magnitude more markets than we currently have, so won't be an issue for a while.

packages/periphery/contracts/Splitter/interfaces/IFeeCoordinator.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.24;
```
EdNoepel, line 2: Recommend prefixing the version with a caret in both interfaces to be flexible with consumers.

packages/periphery/test/unit/Splitter/Splitter.ts
EdNoepel, line 1: nit: file should be named `Splitter.test.ts`

EdNoepel (commented):
No issues with the implementation.
- Would be helpful to update the README for this package with a description of these facilities, as we have for collateral accounts and trigger orders.
- Code documentation is in implementation rather than interfaces.  We agreed to move towards documenting the interfaces and as those are more visible to consumers, and use `@inheritdoc` tags where appropriate.
- Some other comments inline.

packages/periphery/contracts/Splitter/FeeCoordinator.sol
kbrizzle: correct yeah

packages/periphery/contracts/Splitter/FeeSplitter.sol
kbrizzle: We'll need a new setup for V3, so we won't hit any sort of market length constraint here in the meantime.

kbrizzle:
Add to the README and move the natpsec as well.

packages/periphery/contracts/Splitter/FeeSplitter.sol
```diff
@@ -0,0 +1,92 @@
...
+        if (totalSplit.gt(UFixed6Lib.ONE)) revert FeeSplitterOverflowError();
+
+        emit SplitUpdated(beneficiary_, newSplit);
+    }
+
+    /// @inheritdoc IFeeSplitter
+    function poke() external {
+        address[] memory markets = IFeeCoordinator(address(factory())).markets();
+        for (uint256 i; i < markets.length; i++) {
+            IMarket(markets[i]).claimFee(address(this));
```
prateekdefi, line 82: I think we can check if claimable is zero or not before calling `claimFee`. Claim fee only if claimable is non zero to save gas.

EdNoepel (approved):
Looks great!

packages/periphery/contracts/Splitter/FeeSplitter.sol
kbrizzle: I don't think this would save much gas since `claimFee()` internally is pretty much doing this already (read storage slot --> if claimable is zero, don't do anything else).

The second call would also be quite expensive since this is behind a beacon proxy setup.

PR #613, opened 2025-05-27
add script to deploy fee splitter

prateekdefi:
Issue: https://linear.app/perennial/issue/PE-2670/deploy-fee-splitters


packages/deploy/deploy/010_deploy_fee_spliters.ts
arjun-io, line 1: nit: typo in file name (spliters -> splitters)

arjun-io (commented):
Oh, we should also register all markets with the deployed coordinator, market addresses are here: https://github.com/equilibria-xyz/perennial-v2-sdk-ts/blob/5f5c438a4feb57216e70ef4a4b5906f832c9db4b/src/constants/markets.ts#L345

kbrizzle (dismissed):

